PR #6, opened 2026-02-04
chore(test): add nix check and ci tests

gmelodie:
- `nix check`
- CI
- `tests/`: currently just instantiation, `foo` and `getValue`
- better signaling

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,9 +63,5 @@ jobs:
       - name: Run flake checks
         shell: bash
         run: |
-          if [ "${{ runner.os }}" = "macOS" ]; then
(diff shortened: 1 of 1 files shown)


src/libp2p_module_plugin.cpp
```diff
@@ -57,15 +71,16 @@ void Libp2pModulePlugin::peersCallback(
     if (!self) return;
 
     QString message = QString::fromUtf8(msg, int(len));
-
-    // TODO: return peers
+    QString caller = self->lastCaller;
+    QString reqId = QUuid::createUuid().toString();
```
richard-ramos: Returning the peerIds is still a TODO?

PR #15, opened 2026-02-18
feat: stream functions

gmelodie:


tests/async.cpp
```diff
@@ -171,6 +171,7 @@ private slots:
 
         QString uuid = plugin.connectPeer(fakePeer, fakeAddrs);
         auto res = waitForUuid(plugin, *spy, uuid, "connectPeer");
+
```
richard-ramos: The tests are not really testing the stream feature except for checking if it fails if the stream does not exist, right?

richard-ramos (commented):
.

PR #17, opened 2026-02-19
docs: README and kad example

gmelodie:


src/libp2p_module_interface.h
```diff
@@ -7,76 +7,156 @@
...
+    /// Disconnect from a peer.
     Q_INVOKABLE virtual QString disconnectPeer(const QString &peerId) = 0;
+
+    /// Returns information about the local peer.
     Q_INVOKABLE virtual QString peerInfo() = 0;
+
+    /// Returns currently connected peers.
     Q_INVOKABLE virtual QString connectedPeers(int direction = 0) = 0;
+
+    /// Opens a stream using a protocol.
```
richard-ramos: Documentation should indicate that this function returns a streamId which should be used later in the stream functions, and that the stream needs to be released

PR #18, opened 2026-02-19
feat: add mix functions

gmelodie:
- fix current compilation errors in master
- add mix functions
- add privkey generation
- add mix async and sync tests

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,42 +1,43 @@
-add_executable(example_kademlia
-    kademlia.cpp
-)
+# Ensure Qt usage requirements are available in this directory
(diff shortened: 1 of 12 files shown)


src/mix.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include <QtCore/QJsonDocument>
+#include <QtCore/QJsonObject>
+#include <cstring>
+
+QByteArray Libp2pModulePlugin::mixGeneratePrivKey()
+{
+    if (!ctx)
+        return {};
+
+    QString uuid = QUuid::createUuid().toString();
```
richard-ramos: Seems unused

PR #43, opened 2026-05-07
feat: add peerstore

gmelodie:


src/callbacks.cpp
```diff
@@ -35,6 +35,42 @@ void Libp2pModuleImpl::promisePeerInfoCallback(
...
+    int ret, const Libp2pPeerStoreEntry* entry,
+    const char* msg, size_t len, void* userData)
+{
+    auto* p = static_cast<SyncPromise*>(userData);
+    SyncResult r;
+    r.ok = (ret == RET_OK);
+
+    if (ret == RET_OK && entry) {
+        json j;
+        j["peerId"] = entry->peerId ? entry->peerId : "";
```
richard-ramos, line 48: should the publicKey be included too in j? or not really useful?

richard-ramos (approved):

src/callbacks.cpp
gmelodie: Yes! Everything that `entry` has should be in the json

PR #45, opened 2026-05-07
feat: custom protocol registration

gmelodie:
Add custom protocol registration

### DIFF
--- a/src/custom_handlers.cpp
+++ b/src/custom_handlers.cpp
@@ -6,10 +6,6 @@ using json = nlohmann::json;
 // Custom Protocol Handlers
 // ---------------------------------------------------------------------------
 
-static void mountProtocolResultCallback(int ret, const char* msg, size_t len, void* userData) {
-    (void)ret; (void)msg; (void)len; (void)userData;
-}
-
 void Libp2pModuleImpl::protocolHandler(
     libp2p_ctx_t* /*ctx*/, libp2p_stream_t* stream,
     const char* proto, size_t protoLen, void* userData)
@@ -18,7 +14,9 @@ void Libp2pModuleImpl::protocolHandler(
     if (!handlerCtx || !handlerCtx->instance || !stream) return;
 
     auto* self = handlerCtx->instance;
-    std::string protoStr = (proto && protoLen > 0) ? std::string(proto, protoLen) : std::string();
+    std::string protoStr = (proto && protoLen > 0)
+        ? std::string(proto, protoLen)
+        : handlerCtx->proto;
 
     uint64_t streamId = self->addStream(stream);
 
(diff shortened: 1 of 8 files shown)


src/custom_handlers.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+    int ret = libp2p_mount_protocol(
+        ctx, proto.c_str(),
+        &Libp2pModuleImpl::protocolHandler,
+        &mountProtocolResultCallback,
+        handlerCtx.get());
+
+    if (ret != RET_OK) {
+        return {false, {}, "Failed to mount protocol"};
+    }
```
richard-ramos, line 63: This code is assuming that `libp2p_mount_protocol` is sync, but this is not the case. Similar to other functions, libp2p_mount_protocol returns a RET_OK if the function request was enqueued succesfully, but the actual result of the request will be received via callback execution

src/stream.cpp
```diff
@@ -64,8 +64,8 @@ StdLogosResult Libp2pModuleImpl::streamCloseWithEOF(uint64_t streamId) {
 
     auto* p = new SyncPromise();
     auto f = p->get_future();
-    int ret = libp2p_stream_closeWithEOF(ctx, stream,
-                                         &Libp2pModuleImpl::promiseCallback, p);
+    int ret = libp2p_stream_close(ctx, stream,
```
richard-ramos: Why change this from `closeWithEOF` to `close`? at this point this is doing practically the same as the function in line 40.

src/custom_handlers.cpp
```diff
@@ -0,0 +1,51 @@
...
+    uint64_t streamId = self->addStream(stream);
+
+    json j;
+    j["streamId"] = streamId;
+    j["proto"] = protoStr;
+    self->emitEventSafe("protocolStream", j.dump());
+}
+
+StdLogosResult Libp2pModuleImpl::mountProtocol(const std::string& proto) {
+    if (!ctx || !m_started) return {false, {}, "No libp2p context"};
```
richard-ramos: You can mount protocol before they start (see the echo example)

```suggestion
    if (!ctx) return {false, {}, "No libp2p context"};
```

src/plugin.cpp
```diff
@@ -182,6 +181,7 @@ StdLogosResult Libp2pModuleImpl::start() {
 
     auto r = awaitResult(f);
     if (!r.ok) return {false, {}, r.message};
+    m_started = true;
```
richard-ramos, line 185: Set `m_started = false;` in `stop()`

src/custom_handlers.cpp
```diff
@@ -0,0 +1,51 @@
...
+    libp2p_ctx_t* /*ctx*/, libp2p_stream_t* stream,
+    const char* proto, size_t protoLen, void* userData)
+{
+    auto* handlerCtx = static_cast<ProtocolHandlerCtx*>(userData);
+    if (!handlerCtx || !handlerCtx->instance || !stream) return;
+
+    auto* self = handlerCtx->instance;
+    std::string protoStr = (proto && protoLen > 0) ? std::string(proto, protoLen) : std::string();
+
+    uint64_t streamId = self->addStream(stream);
```
richard-ramos, line 21: If emitEvent is nil, this will add the stream to `m_streams` but streamId wont be received by anyone, so, it wont be used neither closed/released thus leaking.

So it's probably a good idea to assert that emitEvent is set (perhaps when mounting the protocol itself?)

richard-ramos (approved):

PR #48, opened 2026-06-10
feat(ci): ASan and TSan jobs for unit tests

gmelodie:
## Summary

Adds ASan and TSan CI coverage for the unit tests. CI previously ran only `nix flake check` with no instrumentation, so wrapper bugs (raw `new`/`delete`, raw pointers shared across the worker thread, "contexts live forever" lifetime model) would not fail the build.

### Changes

**`flake.nix`** — adds `unit-tests-asan` and `unit-tests-tsan` packages per system. Built via `mkLogosModuleTests` with sanitizer flags and runtime opts injected via `preBuild` (env vars: `CC`/`CXX`/`CFLAGS`/`CXXFLAGS`/`LDFLAGS`/`ASAN_OPTIONS`/`TSAN_OPTIONS`). Clang is used as the compiler for better sanitizer diagnostics. `hardeningDisable = ["all"]` to drop Nix stdenv fortify (conflicts with sanitizers). `llvm` added to `nativeBuildInputs` so `llvm-symbolizer` resolves frames to function names — symbol-based suppressions can't match raw hex offsets.

Curated suppression files cover known-noise sources without hiding wrapper bugs:
- **LSan** (`lsan.supp`): suppresses `libp2p.so` (uninstrumented Nim runtime), `std::promise` / `std::__future_base` (the wrapper allocates a `SyncPromise` per async libp2p call; tests that exit while callbacks are in flight leave them undeleted — not a real bug), Qt singletons (`QCoreApplication`, `QMetaObject`, `QArrayData`, `QObjectPrivate`), and `LogosTestRunner` / `LogosTestContext`. Wrapper code (`Libp2pModuleImpl`, `src/*.cpp`) is NOT suppressed — the privkey leak at `src/plugin.cpp:121` still surfaces.
- **TSan** (`tsan.supp`): suppresses `libp2p.so` (Nim runtime races on glibc's non-thread-safe `localtime`). Wrapper-only races (`m_subscribeContexts` in `src/gossipsub.cpp:48-57`) have all-C++ stacks and are unaffected.

**`.github/workflows/ci.yml`**:
- New `sanitizers` matrix job (`[asan, tsan]`) on `ubuntu-latest`, runs `nix build .#unit-tests-<sanitizer>`. Linux x86_64 only.
- Dropped the `cachix-action` step pointing at `name: nix-community` — that cache isn't ours. `magic-nix-cache-action` still handles per-repo caching.

`libp2p.so` itself isn't sanitizer-instrumented, so bugs inside the Nim library are out of scope.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,16 +8,18 @@
 
   outputs = inputs@{ logos-module-builder, ... }:
     let
+      externalLibInputs = {
+        libp2p = {
+          input = inputs.libp2p;
+          packages.default = "cbind";
+        };
+      };
+
       module = logos-module-builder.lib.mkLogosModule {
         src = ./.;
         configFile = ./metadata.json;
         flakeInputs = inputs;
-        externalLibInputs = {
-          libp2p = {
-            input = inputs.libp2p;
-            packages.default = "cbind";
-          };
-        };
+        inherit externalLibInputs;
         tests = {
           dir = ./tests;
         };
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -26,6 +28,70 @@
...
+          clang = pkgs.clang;
+          llvm = pkgs.llvm;  # llvm-symbolizer; symbol-based suppressions need it
+          sanFlags = "-fsanitize=${sanitizer} -fno-omit-frame-pointer -g -O1";
+          tsanSuppressions = pkgs.writeText "tsan.supp" ''
+            race:libp2p.so
+          '';
+          # Wrapper code (Libp2pModuleImpl, src/*.cpp) is NOT suppressed —
+          # real wrapper leaks (e.g. src/plugin.cpp:121 privkey) still surface.
+          lsanSuppressions = pkgs.writeText "lsan.supp" ''
+            leak:libp2p.so
```
richard-ramos, line 54: I read a bit about what these are, and i wonder if we should be more specific with lines 49 and 54? because this will suppress not only libp2p but any races that happen in the C++ code if libp2p appears in the call stack

richard-ramos (approved):

PR #50, opened 2026-06-10
fix: small improvements

gmelodie:
## Summary

A grab-bag of correctness fixes and a deduplication pass over the libp2p cbinding wrapper.

### Correctness

- **Gossipsub unsubscribe was unsubscribing the wrong topic.** `SubscribeCtx` didn't carry the topic, so `gossipsubUnsubscribe` always grabbed `m_subscribeContexts.back()`. Now keyed by topic and guarded by `m_subscribeContextsLock`.
- **Use-after-free in the stream registry.** `getStream` returned a raw `libp2p_stream_t*` under a shared lock, then dropped the lock before issuing an async call that could await up to 10s. A concurrent `streamRelease` would free the pointee. Stream entries now live in a `shared_ptr<StreamEntry>` with a per-stream `shared_mutex` and a `released` flag — read/write/close take shared, release takes exclusive and flips the flag so waiters bail before touching freed memory.
- **Private key heap leak.** Constructor `malloc`d a duplicate of `m_privKey` and never `free`d it. `priv_key.data` now points into `m_privKey.data()` directly; no copy, no leak.
- **Constructor discarded its async ack.** `f.wait_for(5000ms)` returned a status that was never read; `libp2p_new`'s post-callback was a black hole. Now uses `awaitResult` and surfaces failures via `fprintf`.
- **Gossipsub subscribe/unsubscribe async results were silently dropped.** `gossipsubResultCallback` was a `(void)…` black hole. Now emits a `gossipsubResult` event with `{topic, result, message}`.
- **`m_topicQueues[topic]` grew unboundedly.** Polling any topic created an empty queue entry. `gossipsubNextMessage` uses `.find()` now.
- **Missing lock on `m_subscribeContexts`.** Its sibling `m_protocolHandlerContexts` had `m_protocolHandlersLock`; this one had nothing despite the same worker-thread visibility hazard. Added.

### Cleanup

- **Three write-only atomics** (`m_started`, `m_newDone`, `m_destroyDone`) — assigned, never read. Deleted.
- **`mountMix` option** silently ignored since the Mix code was extracted. Removed from `Libp2pModuleOptions`; the two dead references in `examples/mix.cpp` and `tests/integration_mix.cpp` (both excluded from the build) were stripped too.
- **Stale README** advertised "Mix operaions" and "Gossipsub operaions" — Mix is `#if 0`'d in `plugin.h` and the typo was twice. Mix line dropped; typo fixed.
- **`json::parse` exception escape.** 10 sites parsed cbinding output with the throwing overload. Centralized into `parseJsonResponse` using `parse(..., nullptr, false)` + `is_discarded()`.
- **`(ret=N)` interpolation** in every "Failed to X" error message so the diagnostic isn't thrown away at the moment it'\''s most useful.
- **`callSync` / `callSyncWith` / `callSyncStream` template helpers** in `plugin.h` collapse the ~25 copies of `new SyncPromise → invoke → delete-on-fail → awaitResult → check` boilerplate to one or two lines per op across `plugin.cpp`, `stream.cpp`, `kademlia.cpp`, `peerstore.cpp`, `service_discovery.cpp`, `gossipsub.cpp`. The stream variant takes the per-entry shared lock for the duration of the call.
- **Destructor parallelization.** `~Libp2pModuleImpl` was releasing streams serially; with each `streamRelease` awaiting up to 10s, teardown was O(N×10s). Now fans out across `std::async`.
- **`const_cast` on stream writes** dropped — cbinding'\''s `libp2p_stream_write`/`writeLp` take `const uint8_t*`. Kept (with a comment) on `gossipsub_publish` since that cbinding signature is still non-const.

### DIFF
--- a/src/gossipsub.cpp
+++ b/src/gossipsub.cpp
@@ -1,5 +1,6 @@
 #include "plugin.h"
 
+#include <algorithm>
 #include <cstring>
 
 // ---------------------------------------------------------------------------
@@ -25,8 +26,15 @@ StdLogosResult Libp2pModuleImpl::gossipsubPublish(
     return {true, {}, ""};
 }
 
+// Surface async (un)subscribe outcomes as gossipsubResult events.
 static void gossipsubResultCallback(int ret, const char* msg, size_t len, void* userData) {
-    (void)ret; (void)msg; (void)len; (void)userData;
+    auto* subCtx = static_cast<Libp2pModuleImpl::SubscribeCtx*>(userData);
+    if (!subCtx || !subCtx->instance) return;
+    nlohmann::json j;
+    j["topic"] = subCtx->topic;
+    j["result"] = ret;
+    j["message"] = (msg && len > 0) ? std::string(msg, len) : std::string();
+    subCtx->instance->emitEventSafe("gossipsubResult", j.dump());
 }
 
 StdLogosResult Libp2pModuleImpl::gossipsubSubscribe(const std::string& topic) {
@@ -34,6 +42,7 @@ StdLogosResult Libp2pModuleImpl::gossipsubSubscribe(const std::string& topic) {
 
     auto subCtx = std::make_unique<SubscribeCtx>();
     subCtx->instance = this;
+    subCtx->topic = topic;
 
     int ret = libp2p_gossipsub_subscribe(
         ctx, topic.c_str(),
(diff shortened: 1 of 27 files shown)


src/stream.cpp
```diff
@@ -2,165 +2,107 @@
...
-    int ret = libp2p_stream_release(ctx, stream,
+    int ret = libp2p_stream_release(ctx, entry->ptr,
                                     &Libp2pModuleImpl::promiseCallback, p);
-    if (ret != RET_OK) { delete p; return {false, {}, "Failed to release stream"}; }
-
-    removeStream(streamId);
+    if (ret != RET_OK) {
+        delete p;
+        return {false, {}, "Failed to release stream (ret=" + std::to_string(ret) + ")"};
+    }
```
richard-ramos: `removeStream` should be called *after* `libp2p_stream_release` releases the stream right?

src/plugin.cpp
```diff
@@ -118,138 +105,105 @@ Libp2pModuleImpl::Libp2pModuleImpl(const Libp2pModuleOptions& options)
...
         for (const auto& [id, _] : m_streams) {
             streamIds.push_back(id);
         }
     }
+    std::vector<std::future<StdLogosResult>> releases;
+    releases.reserve(streamIds.size());
     for (uint64_t sid : streamIds) {
-        streamRelease(sid);
+        releases.push_back(std::async(std::launch::async,
+                                      [this, sid] { return streamRelease(sid); }));
```
richard-ramos: Could we use a worker pool to limit the maximum number of threads we are creating here?

src/plugin.cpp
```diff
@@ -118,138 +105,105 @@ Libp2pModuleImpl::Libp2pModuleImpl(const Libp2pModuleOptions& options)
...
+        streamIds.reserve(m_streams.size());
         for (const auto& [id, _] : m_streams) {
             streamIds.push_back(id);
         }
     }
+    std::vector<std::future<StdLogosResult>> releases;
+    releases.reserve(streamIds.size());
     for (uint64_t sid : streamIds) {
-        streamRelease(sid);
+        releases.push_back(std::async(std::launch::async,
```
richard-ramos: Async operations can throw, so we should wrap this in a try/catch since raising exceptions inside a destructor are not allowed

src/gossipsub.cpp
```diff
@@ -1,92 +1,103 @@
...
+    // Ctx lives until destructor — the async result callback still uses it.
+    SubscribeCtx* ctxPtr = nullptr;
+    {
+        std::lock_guard<std::mutex> lock(m_subscribeContextsLock);
+        auto it = std::find_if(m_subscribeContexts.begin(), m_subscribeContexts.end(),
+            [&](const std::unique_ptr<SubscribeCtx>& c) { return c->topic == topic; });
+        if (it == m_subscribeContexts.end()) {
+            return {false, {}, "Not subscribed to topic"};
+        }
+        ctxPtr = it->get();
```
richard-ramos, line 82: We should also remove this `SubscribeCtx` once `libp2p_gossipsub_unsubcribe` is succesful, otherwise we kept a stale value inside `m_subscribeContexts`

richard-ramos (approved):

PR #55, opened 2026-06-17
chore: extract mix bindings + bump logos-module-builder

gmelodie:
## Mix extraction
Removes the mix bindings that were already disabled (commented out / `#if 0`) and have been extracted to a separate repo with no cbindings yet:
- Deletes `src/mix.cpp`, `examples/mix.cpp`, `tests/integration_mix.cpp`
- Drops the leftover commented `mix` references in `CMakeLists.txt`, `examples/CMakeLists.txt`, `tests/CMakeLists.txt`
- Strips the `#if 0` mix blocks from `src/plugin.h` and `tests/sync.cpp`

## Bump nim-libp2p
Bumps `nim-libp2p` (`3f5a4dc` → `9d38a6c`) in `flake.lock`. Knock-on changes from the new FFI:
- Widens integer parameters to fixed-width types so codegen keeps them: `int` → `int64_t` (`connectedPeers`, `gossipsubNextMessage`, `kadGetValue`) and `size_t` → `uint64_t` (`streamReadExactly`, `streamReadLp`) in `src/plugin.h` and the matching `.cpp` files.
- Removes default arguments from `plugin.h` method declarations (`connectPeer`, `discoStartAdvertising`, `discoLookup`, `peerstoreAddPeer`, `circuitRelayReserve`, …) and collapses them to single-line signatures, both required for codegen to emit them.
- `src/plugin.cpp`: nim-libp2p no longer supplies an implicit listen address, so we fall back to a localhost address matching the selected transport (`/ip4/127.0.0.1/udp/0/quic-v1` for QUIC, `/ip4/127.0.0.1/tcp/0` otherwise) when none is given.
- Updates `tests/integration_service_discovery.cpp` to pass the now-required `serviceData` argument explicitly.

## Deps
- Un-forks `logos-module-builder` to mainline (`fix/codegen-explicit-ctor` → pinned rev `62524be` on `github:logos-co/logos-module-builder`) and updates `flake.lock` accordingly.

## CI
- `tests/integration_e2e/openmetrics_e2e.sh`: on a `wait_until` timeout, dump the last attempt's output and tail the daemon log to make e2e failures debuggable.

#52 is rebased on top of this branch.

PR #56, opened 2026-06-17
feat(openmetrics): libp2p metrics

gmelodie:
## Summary

Finishes OpenMetrics support by returning **real** values from nim-libp2p instead of the placeholder zeros, on top of the mix/deps prep extracted into #55.

`collectMetrics()` now calls the new `libp2p_collect_metrics` cbinding and **passes the entire nim-libp2p metrics registry through** — one OpenMetrics entry per series, with labels preserved (no summing). `type` and `help` are derived from the `# TYPE` / `# HELP` lines in the exposition. Before the node is started (no `ctx`) or on any FFI error the call fails, the text stays empty, and an empty-but-well-formed `metrics` array is returned.

This replaces the earlier fixed 7-series allowlist: the exposed schema now follows whatever nim-libp2p records (connection, dial, pubsub, gossipsub scoring, service-discovery `cd_*`, etc.), so labelled per-peer / per-topic series come through individually rather than being collapsed.

Rendering details handled by the parser:
- **Counters** keep their nim-metrics `_total` series; the `_created` timestamp series each counter/histogram emits is dropped (metadata, not a metric value).
- **Histograms / summaries** pass through as their `_bucket` / `_sum` / `_count` (and `quantile`) component series, typed from the family's `# TYPE` line.
- **Labels** are parsed into the entry's `labels` object, honouring quoted values that contain commas or escaped quotes.

## Base

Targets `chore/extract-mix` (#55), which carries the mix-binding removal and `logos-module-builder` un-fork. This PR is the OpenMetrics-only slice of the now-closed #52.

## Requires

Depends on the cbind export in vacp2p/nim-libp2p#2625 (`libp2p_collect_metrics` + building with `-d:metrics`). The `libp2p` flake input is bumped to that branch/rev; re-point it at `master` once #2625 merges.

## Note

The exposed metric set is now dynamic (it mirrors nim-libp2p's registry) rather than a curated fixed schema. Downstream consumers should not assume a fixed series list.


src/metrics.cpp
```diff
@@ -3,33 +3,28 @@
...
+            return {true, r.message, ""};
         });
+    if (res.success && res.value.is_string()) {
+        text = res.value.get<std::string>();
+    }
+
+    const auto parsed = json::parse(text, nullptr, false);
+    std::vector<Metric> series;
+    if (parsed.is_array()) {
+        series = parsed.get<std::vector<Metric>>();
```
richard-ramos:
```suggestion
        try {
            series = parsed.get<std::vector<Metric>>();
        } catch (const nlohmann::json::exception&) {
            series.clear();
        }
```

src/plugin.h
```diff
@@ -68,6 +69,50 @@ inline StdLogosResult parseJsonResponse(const std::string& s, const char* errPre
...
+inline void to_json(nlohmann::json& j, const Metric& m) {
+    j = nlohmann::json{
+        {"name", m.name},
+        {"type", m.type},
+        {"help", m.help},
+        {"labels", m.labels},
+        {"value", m.value},
+    };
+    if (m.timestamp != 0.0) j["timestamp"] = m.timestamp;
+}
```
richard-ramos: Is timestamp a float or unix milisecond

src/metrics.cpp
```diff
@@ -3,33 +3,28 @@
...
-            {"type",  s.type},
-            {"help",  s.help},
-            {"value", 0},
+    // Pass through nim-libp2p's whole metrics registry. It returns a JSON array
+    // of metric series; before start (no ctx) or on error the call fails and we
+    // return an empty-but-well-formed metrics array.
+    std::string text;
+    auto res = callSyncWith("Failed to collect metrics",
+        [&](SyncPromise* p) {
+            return libp2p_collect_metrics(ctx, &Libp2pModuleImpl::promiseCallback, p);
```
richard-ramos, line 12: It's possible to get a double free when using `callSyncWith` and `promiseCallback`, because both delete `p` if `libp2p_collect_metrics` fails, executing  `promiseCallback` and passing `RET_ERR`; and then, `callSyncWith` will see the `ret != RET_OK` and try to `delete p` again

richard-ramos (approved):

PR #59, opened 2026-06-22
feat: multi-node bootstrap service discovery example

gmelodie:
## Summary

Two related changes:

1. **Service Discovery example** now demonstrates real cross-node discovery. It went from two directly-connected nodes (where `discoLookup` returned `[]` and looked broken) to a three-node bootstrap topology — a bootstrap node anchors the DHT, an advertiser and a discoverer bootstrap off it, and the discoverer resolves the advertiser through the DHT.
2. **Module config plumbing** so a `logoscore`-loaded module can be pointed at listen addrs / bootstrap nodes. Since loaded modules are default-constructed, options come from the `LIBP2P_MODULE_CONFIG` env var (inline JSON or a path to a JSON file), overlaid onto the defaults at construction.

## Changes

- `examples/service_discovery.cpp` — three-node bootstrap topology; lookup polls until a peer appears instead of a fixed sleep; soft-reports match/no-match (exit 0) so it stays a teaching example.
- `src/plugin.h` — `Libp2pModuleOptions::load()` reads `LIBP2P_MODULE_CONFIG` and overlays it onto the defaults. Unset / unreadable / invalid-JSON / wrong-typed field all log a warning and fall back to defaults (never throws out of construction).
- `metadata.json` / `README.md` / `examples/README.md` — document the env var, schema, and the new example.
- `tests/config.cpp` — overlay, partial config, env inline/file, and the fallback cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,35 @@ Check the examples/ directory for complete usage demonstrations.
 
 ---
 
+# Configuration
+
+When the module is loaded by `logoscore`, it is constructed without arguments, so
+options are supplied through the `LIBP2P_MODULE_CONFIG` environment variable — set
+to either inline JSON or a path to a JSON file. The config is overlaid onto the
+defaults at load time; every key is optional, and an omitted key keeps its default.
+If the config is unset, unreadable, not valid JSON, or has a wrong-typed field, the
+module logs a warning and falls back to the full default options.
+
+```bash
+# inline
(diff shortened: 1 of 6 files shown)


jm-clius (approved):
Yes, this seems to me the missing part of the puzzle for end-to-end use of the service discovery module. 
Thanks!

PR #62, opened 2026-06-23
feat: user-provided XPRs

gmelodie:
## Summary

Exposes a `createXpr` method on the libp2p module so callers can build and sign the node's own Extended Peer Record (XPR) on demand, rather than relying solely on the records produced internally by service discovery.

This is the C-binding side of the change; it points the `libp2p` flake input at the matching `feat/cbind/user-provided-xprs` branch of `nim-libp2p`, which provides the underlying `libp2p_create_xpr` FFI entry point.

## Changes

- `src/plugin.h` / `src/service_discovery.cpp`: add `createXpr(addrs, services, seqNo)`. It marshals the service list into `Libp2pServiceInfo` and calls `libp2p_create_xpr`, returning the signed protobuf bytes. Empty `addrs` falls back to the node's listen addresses; a `seqNo` of `0` defaults to the current time.
- `src/plugin.h`: extract the repeated buffer→result conversion into a shared `bufferToResult` helper, and route `createXpr`, `publicKey`, `newPrivateKey`, `kadGetValue`, and the stream reads through it (previously six copies of the same line).
- `tests/integration_service_discovery.cpp`: add the `create_xpr` integration test, covering the explicit-addrs path, the empty-addrs/`seqNo`-0 default path, and a seqNo-only change that isolates the seqNo branch and asserts it alters the signed bytes.
- `examples/service_discovery.cpp` / `examples/README.md`: demonstrate building a signed XPR from the advertiser node, print its size, surface the failure path, and document the new capability.
- `flake.nix` / `flake.lock`: bump the `libp2p` input from `chore/cbind/metrics` to `feat/cbind/user-provided-xprs`.

## On the returned value

The XPR result is raw signed protobuf bytes — opaque binary, not UTF-8 text. It is carried as a `std::string` used purely as a byte container, the same convention every other binary-returning method uses (`publicKey`, `newPrivateKey`, `kadGetValue`, stream reads), now centralized in `bufferToResult`. Callers should treat it as bytes; the example prints only `.size()`, never the content.

## Notes

Depends on the `nim-libp2p` `feat/cbind/user-provided-xprs` branch being merged; the flake input must be re-pointed to its final ref before this lands.

### DIFF
--- a/src/service_discovery.cpp
+++ b/src/service_discovery.cpp
@@ -82,9 +82,8 @@ StdLogosResult Libp2pModuleImpl::discoRandomLookup() {
         });
 }
 
-/// Builds and signs the node's own Extended Peer Record (XPR) with its private
-/// key, returning the protobuf-encoded signed bytes. Empty `addrs` falls back
-/// to the node's listen addresses; `seqNo` of 0 defaults to the current time.
+/// Builds and signs the node's own Extended Peer Record, returning the signed
+/// protobuf bytes. Empty `addrs` uses the listen addresses; `seqNo` 0 uses now.
 StdLogosResult Libp2pModuleImpl::createXpr(
     const std::vector<std::string>& addrs,
     const std::vector<std::pair<std::string, std::string>>& services,
@@ -112,7 +111,5 @@ StdLogosResult Libp2pModuleImpl::createXpr(
                 serviceInfos.empty() ? nullptr : serviceInfos.data(), serviceInfos.size(),
                 seqNo, &Libp2pModuleImpl::promiseBufferCallback, p);
         },
-        [](const SyncResult& r) -> StdLogosResult {
-            return {true, std::string(r.buffer.begin(), r.buffer.end()), ""};
-        });
+        bufferToResult);
 }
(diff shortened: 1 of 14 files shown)


src/service_discovery.cpp
```diff
@@ -81,3 +81,38 @@ StdLogosResult Libp2pModuleImpl::discoRandomLookup() {
...
+
+    return callSyncWith("Failed to create XPR",
+        [&](SyncPromise* p) {
+            return libp2p_create_xpr(
+                ctx, addrPtrs.empty() ? nullptr : addrPtrs.data(), addrPtrs.size(),
+                serviceInfos.empty() ? nullptr : serviceInfos.data(), serviceInfos.size(),
+                seqNo, &Libp2pModuleImpl::promiseBufferCallback, p);
+        },
+        [](const SyncResult& r) -> StdLogosResult {
+            return {true, std::string(r.buffer.begin(), r.buffer.end()), ""};
```
richard-ramos: What format is this using to print the buffer. Is it safe to assume that this is UTF8 string?

richard-ramos (approved):

PR #64, opened 2026-06-25
feat(config): configurable persistent private key

gmelodie:
## Summary

The node generated a fresh `secp256k1` key on every startup, so its peer ID changed on every restart. For a DHT/bootstrap node — whose whole job is being a stable, dialable address other peers pin in their routing tables — a peer ID that churns on restart is a real operational limitation.

This adds two optional config keys (and matching `Libp2pModuleOptions` fields), so an operator can supply and persist an identity:

- `privKey` — hex-encoded raw private key bytes. When set, the node reuses it instead of generating one, giving a stable peer ID across restarts.
- `keyType` — `"rsa" | "ed25519" | "secp256k1" | "ecdsa"`, used only when generating a fresh key (i.e. `privKey` omitted). Defaults to `"secp256k1"`, preserving the previous behavior.

## Changes

- `Libp2pModuleOptions` gains `privKey` and `keyType`; the node uses a supplied key when present and otherwise generates one of `keyType`.
- `newPrivateKey()` now delegates to a new private `generatePrivateKey(scheme)` helper that context creation also uses. Its return encoding changed from base64 to **hex**, so its output is directly usable as the `privKey` config value (round-trips through `decodeHex`). This is a behavior change for any caller that previously base64-decoded the result.
- Config parsing: `parseKeyType` maps the string to the scheme; `decodeHex` parses `privKey`, throwing `std::invalid_argument` on a non-string, odd-length, or non-hex value so `load()` falls back to defaults (consistent with how wrong-typed fields are already handled). No key material is included in error messages.
- The persistent `m_privKey` is held in a `SecureBytes` (a `std::vector` backed by a `ScrubbingAllocator`) so the key bytes are wiped from the heap on reallocation and destruction; the `volatile` writes are not elided by the optimizer the way a plain `memset` would be.
- Shared helpers live in `utils` (`utils.h`/`utils.cpp`): the hex/base64 codecs (`hexEncode`, `decodeHex`, `base64Encode`) and the `ScrubbingAllocator`/`SecureBytes` facility, rather than being re-implemented per call site.
- Docs: `metadata.json` schema and `README.md` example document both keys.

## Tests

`tests/config.cpp` adds:
- `config_from_json_key_type_and_priv_key` — `keyType`/`privKey` parse correctly (incl. mixed-case hex).
- `config_key_type_defaults_to_secp256k1` — defaults preserved.
- `config_load_invalid_priv_key_hex_returns_defaults` / `config_load_odd_priv_key_hex_returns_defaults` — malformed hex falls back to defaults without throwing out of construction.
- `config_priv_key_stable_peer_identity` — a supplied key (the hex from `newPrivateKey()`, decoded via `decodeHex`) yields the same peer ID across two separate constructions.

## Notes

Scrubbing covers the long-lived `m_privKey`. The key still passes transiently through the config string, the cbinding callback buffer, a local during generation, and the un-scrubbed `Libp2pModuleOptions::privKey` field; wiping those would mean threading the secure allocator through the shared promise machinery and the public options struct — left out as high-blast-radius, low-value. `LIBP2P_MODULE_CONFIG` itself remains plaintext by design of the config system, so an operator-supplied `privKey` is readable from the process environment regardless.


src/plugin.h
```diff
@@ -99,6 +138,10 @@ inline void apply(const nlohmann::json& j, Libp2pModuleOptions& o) {
         }
     }
     o.transport = parseTransport(j, o.transport);
+    o.keyType = parseKeyType(j, o.keyType);
+    if (auto it = j.find("privKey"); it != j.end() && it->is_string()) {
```
richard-ramos: Could we throw an exception if the privKey is present but not a string? i.e.:
```
        throw std::invalid_argument("privKey must be a string");
```

src/plugin.cpp
```diff
@@ -217,7 +220,7 @@ SyncResult Libp2pModuleImpl::generatePrivateKeyRaw() {
 }
 
 StdLogosResult Libp2pModuleImpl::newPrivateKey() {
-    auto r = generatePrivateKeyRaw();
+    auto r = generatePrivateKey(LIBP2P_PK_SECP256K1);
     if (!r.ok) return {false, {}, r.message};
     return bufferToResult(r);
```
richard-ramos: This will return a base64 private key, but the config then expects it to be in hex format. To be consistent then the private key here should be returned as a hex string and not as base64.

src/plugin.h
richard-ramos, line 239: This seems to be stale

src/plugin.cpp
richard-ramos, line 167: Something to fix in nim-libp2p:
the code looks like this:
```
if config.privKey.data != nil and config.privKey.dataLen > 0:
  let keySeq = config.privKey.toByteSeq()
  PrivateKey.init(keySeq).withValue(copyKey):
    privKey = Opt.some(copyKey)
```

meaning that if `PrivateKey.init` fails, it will prcoeed, creating a random key, instead of failing, which is what one would expect if you pass an invalid private key.

richard-ramos (approved):

src/plugin.cpp
gmelodie: https://github.com/vacp2p/nim-libp2p/pull/2717

PR #65, opened 2026-06-25
chore: dedup the sync-over-async bridge, split plugin.h, surface init failure

gmelodie:
## Summary

Cleans up the C++ shim that bridges the Logos sync module API to nim-libp2p's async C bindings. No behavioural change for callers on success paths; empty/error results are preserved. Motivation was reducing the maintenance surface of the callback layer and closing a few sharp edges found while reading it.

## Correctness

- Drop a dead `m_libp2pConfig.addrs = m_addrsPtr.data()` in the constructor that ran while `m_addrsPtr` was still empty (it was immediately overwritten by the correct populate-then-assign block, but read as a bug).
- Stop a caller-supplied op timeout from being silently capped by the C++ await: `callSync`/`callSyncWith` now take an `awaitMs`, and `connectPeer` passes `awaitTimeoutFor(timeoutMs)` (op timeout + 5s slack, `INT_MAX`-clamped). Previously a 30s connect returned a spurious `"timeout"` at the hardcoded 10s await.
- Add the missing null guard to `mountCompleteCallback` (`!hCtx || !hCtx->mountPromise`) that every other callback already had.

## Surface init failure

The constructor is default-constructed by codegen and could not signal failure, leaving `ctx == nullptr` indistinguishable from "not started". Added `m_initError`, populated on each failure path (key-gen, `libp2p_new`, null ctx), exposed via two new accessors: `bool ok()` and `StdLogosResult status()`.

## Timeouts

Replaced the magic `10000`/`5000` literals with named constants: `kDefaultOpTimeoutMs`, `kNewContextTimeoutMs`, `kDestroyTimeoutMs`, `kAwaitSlackMs`.

## Dedup and structure

- New `finishPromise()` + `basicResult()` helpers collapse the identical seed/resolve/`delete p` boilerplate repeated across every `promise*Callback`.
- New `jsonResult(r, emptyDefault)` helper replaces the 8 copies of `if (r.message.empty()) ... else parseJsonResponse(...)`.
- Structured callbacks now write a `nlohmann::json` straight into `SyncResult::data` instead of `.dump()`-ing a string that ops re-parsed; the now-unused `parseJsonResponse` is removed.
- Replaced the type-unsafe `void* extra` with a typed `libp2p_stream_t* stream` field (`dial`/`dialCircuitRelay`).
- Split `plugin.h` (418 → 309 lines): config parsing moved to `src/config.h`, `Metric` (de)serialization to `src/metric.h`. `plugin.h` includes both, so the codegen `impl_header` contract and the existing tests are untouched.

## Testing

Could not run the test suite in the sandbox — the Logos test harness pulls in Qt, which isn't available here. All `src/` translation units were syntax-checked against the real `cbind/libp2p.h` and Logos SDK headers, and a probe TU confirms the moved symbols (`libp2p_module_config::apply`, `Libp2pModuleOptions::load`, `Metric`) and the new `ok()`/`status()` accessors resolve through `<plugin.h>`. Needs a CI run for the full integration + sanitizer suite.

### DIFF
--- a/src/plugin.cpp
+++ b/src/plugin.cpp
@@ -119,6 +119,7 @@ Libp2pModuleImpl::Libp2pModuleImpl(const Libp2pModuleOptions& options)
     if (!r.ok) {
         m_initError = "libp2p_new failed: " + r.message;
         fprintf(stderr, "libp2p_new failed: %s\n", r.message.c_str());
+        ctx = nullptr;
     }
 
     if (!ctx) {
--- a/src/plugin.cpp
+++ b/src/plugin.cpp
@@ -17,25 +17,16 @@ std::string defaultListenAddr(int transport) {
 }
 
 void Libp2pModuleImpl::promiseCallback(int ret, const char* msg, size_t len, void* userData) {
-    auto* p = static_cast<SyncPromise*>(userData);
-    SyncResult r;
-    r.ok = (ret == RET_OK);
-    r.message = (msg && len > 0) ? std::string(msg, len) : std::string();
-    p->set_value(std::move(r));
-    delete p;
+    finishPromise(static_cast<SyncPromise*>(userData), basicResult(ret, msg, len));
 }
 
 void Libp2pModuleImpl::promiseBufferCallback(int ret, const uint8_t* data, size_t dataLen,
                                              const char* msg, size_t len, void* userData) {
-    auto* p = static_cast<SyncPromise*>(userData);
-    SyncResult r;
-    r.ok = (ret == RET_OK);
-    r.message = (msg && len > 0) ? std::string(msg, len) : std::string();
+    auto r = basicResult(ret, msg, len);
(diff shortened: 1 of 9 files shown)


src/plugin.cpp
```diff
@@ -195,6 +187,15 @@ StdLogosResult Libp2pModuleImpl::stop() {
     });
 }
 
+bool Libp2pModuleImpl::ok() {
+    return ctx != nullptr;
+}
+
+StdLogosResult Libp2pModuleImpl::status() {
+    if (ctx) return {true, {}, ""};
```
richard-ramos, line 196: This is returning true if ctx is not nil, but this is not necessarily correct, because libp2p_new is async, so it will return the `ctx` immediatly in cbind/libp2p.nim: 293, , but the request could later fail when the request created in cbind/libp2p.nim:285 is processed

PR #66, opened 2026-06-25
fix(tsan): guard emitEvent against the worker-thread data race

gmelodie:
## Summary

`emitEvent` is a public `std::function` set by the host and then invoked from libp2p worker threads (`eventCallback`, `topicHandler`, `protocolHandler`, `gossipsubResultCallback`). It was read without synchronization, so a reassignment after callbacks can fire is a data race — TSAN-visible, and the reason for this branch. This guards the read path.

## What changed

- Added a lock-guarded snapshot `m_emitEventSnapshot` (under `m_emitEventLock`), published on the caller thread via `publishEmitEvent()` at every point a worker can begin emitting: `start()`, `setEventCallback()`, and `mountProtocol()` (`src/plugin.cpp`, `src/custom_handlers.cpp`).
- `emitEventSafe()` now copies the snapshot under a `shared_lock` and invokes it *after* releasing the lock, so a re-entrant callback can't deadlock against `publishEmitEvent()`'s `unique_lock` (`src/plugin.cpp`).
- The public `emitEvent` field is unchanged, so the existing `node.emitEvent = ...` assignment in tests and examples keeps working; worker threads now read only the snapshot, never the field.
- Added an `EmitEventFn` alias to dedupe the repeated function signature (`src/plugin.h`).

## Why a snapshot instead of locking the field directly

The host assigns `emitEvent` directly through the public field, so a write can't take a lock. Snapshotting on the caller thread before any worker starts emitting establishes a clean happens-before: workers read only the guarded copy. The "set once before start" contract still holds; the additional publishes in `setEventCallback`/`mountProtocol` re-snapshot defensively if the host sets it late.

## Testing

The guard pattern was compiled in isolation (`g++ -std=c++17 -Wall -Wextra`) — clean. The full module build requires the libp2p toolchain (`LOGOS_MODULE_BUILDER_ROOT`), which is not available in this environment.

### DIFF
--- a/src/custom_handlers.cpp
+++ b/src/custom_handlers.cpp
@@ -39,6 +39,7 @@ StdLogosResult Libp2pModuleImpl::mountProtocol(const std::string& proto) {
     // Without emitEvent, protocolHandler would register a stream that no caller
     // could ever read, close, or release — leaking the stream.
     if (!emitEvent) return {false, {}, "emitEvent must be set before mounting a protocol"};
+    publishEmitEvent();
 
     auto handlerCtx = std::make_unique<ProtocolHandlerCtx>();
     handlerCtx->instance = this;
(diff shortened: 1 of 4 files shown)


richard-ramos (approved):
Should publishEmitEvent() be called also in libp2p_gossipsub_subscribe and libp2p_gossipsub_unsubscribe ?

PR #71, opened 2026-06-29
feat: run a libp2p node standalone via createNode/getNodeInfo

gmelodie:
## Summary

Before this change there was no way to drive a libp2p node from `logoscore` the way other modules are driven — `libp2p_module` had no `createNode` and no `getNodeInfo`, and configuration came only from the `LIBP2P_MODULE_CONFIG` env var read at load time. This adds the two call-time methods so the module can run standalone end-to-end:

```
logoscore load-module libp2p_module
logoscore call libp2p_module createNode @config.example.json
logoscore call libp2p_module start
logoscore call libp2p_module getNodeInfo Version
logoscore call libp2p_module getNodeInfo MyBoundPorts
```

## Changes

- `src/plugin.{h,cpp}`: extracted the constructor/destructor bodies into reusable `applyOptions()` / `createContext()` / `destroyContext()` helpers, then added two single-line public methods (kept single-line / no default args / fixed-width ints so the codegen impl-header parser exposes them over RPC):
  - `createNode(config)` parses the call-time JSON (same schema as the env var), tears down the auto-created node, and rebuilds from it. Invalid JSON returns an error *before* any teardown, so a bad config leaves the existing node intact.
  - `getNodeInfo(field)` dispatches `Version` (module version constant), `MyBoundPorts` (ports parsed from `peerInfo()` multiaddrs via a new `extractPort` helper), `PeerId`, and `Multiaddrs`; an unknown field returns an error.
- `src/config.h`: added `Libp2pModuleOptions::fromJson(raw, ok, err)` and routed `load()` through it, so the env-var and `createNode` paths share one schema parser. `fromJson` now reports *why* a config was rejected (`malformed JSON` or the offending field) through the optional `err` out-param.
- Better failure reporting: `logoscore` only relays a generic "call failed" to the CLI, so `createNode` now writes the specific reason (`createNode: invalid config: …`) to the daemon's stderr, and the `LIBP2P_MODULE_CONFIG` warning includes the reason too.
- `config.example.json`: a minimal, runnable standalone config referenced from the README.
- Docs: expanded the README "Running a node via logoscore" section into a full build → install → daemon → call walkthrough (including the load-module race and the "errors go to the daemon log, not the CLI" gotcha), plus `metadata.json` and `tests/README.md` notes.

## Backward compatibility

The constructor still auto-creates a node, so `examples/`, the existing C++ tests, and the env-var standalone flow are unchanged. `createNode` is optional — set `LIBP2P_MODULE_CONFIG` before load and you can `start` directly. The new `fromJson` `err` argument defaults to `nullptr`, so existing two-argument callers are unaffected.

## Testing

- Unit tests for `fromJson` in `tests/unit_config.cpp` (valid overlay, invalid JSON, wrong-typed field, and the new `err`-reason reporting).
- Integration tests in `tests/integration.cpp`: `integration_create_node_then_node_info` (rebuild → start → `Version`/`MyBoundPorts`/`PeerId` → unknown-field error) and `integration_create_node_invalid_config_fails` (now asserts the error message).
- A dedicated `tests/integration_e2e/standalone_e2e.sh`, wired as `nix run .#standalone-e2e` and run in CI, drives only `libp2p_module` through real `logoscore` RPC and asserts the command outputs (`Version`, bound port, `PeerId`, `Multiaddrs`) plus the negative paths (unknown field rejected; malformed `createNode` rejected and its reason logged).
- The existing `openmetrics_e2e.sh` still exercises `createNode` + `getNodeInfo Version` over real RPC.

Note: `config.h` was syntax-checked locally; the integration and e2e layers run in CI, where the libp2p `.so` is vendored at build time.

## Notes

`getNodeInfo Version` returns the hardcoded module version (`1.0.0`, mirroring `metadata.json`) because the cbind `libp2p.h` exposes no version symbol. Tracking `metadata.json` automatically or surfacing a nim-libp2p build id can be a follow-up.

### DIFF
--- a/src/plugin.cpp
+++ b/src/plugin.cpp
@@ -161,20 +161,23 @@ StdLogosResult Libp2pModuleImpl::createContext() {
     auto* p = new SyncPromise();
     auto f = p->get_future();
 
-    ctx = libp2p_new(&m_libp2pConfig, &Libp2pModuleImpl::promiseCallback, p);
+    auto* candidate = libp2p_new(&m_libp2pConfig, &Libp2pModuleImpl::promiseCallback, p);
 
     auto r = awaitResult(f, kNewContextTimeoutMs);
     if (!r.ok) {
         m_initError = "libp2p_new failed: " + r.message;
         fprintf(stderr, "libp2p_new failed: %s\n", r.message.c_str());
-        ctx = nullptr;
+        if (candidate) destroyHandle(candidate);
+        return {false, {}, m_initError};
     }
 
-    if (!ctx) {
-        if (m_initError.empty()) m_initError = "libp2p_new returned null context";
+    if (!candidate) {
+        m_initError = "libp2p_new returned null context";
         fprintf(stderr, "libp2p_new returned null context\n");
         return {false, {}, m_initError};
     }
+
+    ctx = candidate;
     return {true, {}, ""};
 }
 
@@ -187,7 +190,11 @@ StdLogosResult Libp2pModuleImpl::createNode(const std::string& config) {
         fprintf(stderr, "libp2p_module: %s\n", msg.c_str());
         return {false, {}, msg};
     }
-    destroyContext();
+    if (ctx) {
+        std::string msg = "createNode: node already created";
+        fprintf(stderr, "libp2p_module: %s\n", msg.c_str());
+        return {false, {}, msg};
+    }
     applyOptions(options);
     return createContext();
 }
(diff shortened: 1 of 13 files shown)


src/plugin.cpp
richard-ramos: Can we keep the returned ctx in a temporary variable until the create callback succeeds and clear it otherwise? because `libp2p_new` returns a context once it succesfully enqueues the request, but then that callback can later fail with RET_ERR, and when that happens this code sets ctx = nullptr in line 170, thus we lose the pointer and cannot call libp2p_destroy() in it

src/plugin.cpp
```diff
@@ -136,54 +173,74 @@ Libp2pModuleImpl::Libp2pModuleImpl(const Libp2pModuleOptions& options)
...
+StdLogosResult Libp2pModuleImpl::createNode(const std::string& config) {
+    bool ok = false;
+    std::string err;
+    auto options = Libp2pModuleOptions::fromJson(config, ok, &err);
+    if (!ok) {
+        std::string msg = "createNode: invalid config: " + err;
+        fprintf(stderr, "libp2p_module: %s\n", msg.c_str());
+        return {false, {}, msg};
+    }
+    destroyContext();
```
richard-ramos: Why destroy the context here? does this mean that you can call createNode(config) more than once?

gmelodie: Yes! Exactly, so `createNode` can be called more than once without leaking

richard-ramos: Then some additional work is needed as m_streams, m_sbuscribeContexts, m_topicQueues and m_protocolHandlerContexts can still contain data from the old node at the moment you call createNode(), so it's necessary to make sure all state is cleaned up when rebuilding the node

gmelodie: added `clearNodeState()`

richard-ramos (approved):

PR #72, opened 2026-06-29
feat: add INFO-level logs for service discovery events

tinniaru3005:
Adds two INFO-level log lines to help measure and trace service discovery:

- `[INFO] <timestamp> service-discovery: starting interest for service '<serviceId>'`  
  logged when `discoRegisterInterest` is called

- `[INFO] <timestamp> service-discovery: found peer '<peerId>' offering service '<serviceId>'`  
  logged when `discoLookup` returns a result with peers

These logs make it possible to measure discovery latency as the time between the two log lines, addressing the discussion around how to track "time to discover" for simulation scenarios.

### DIFF
--- a/src/service_discovery.cpp
+++ b/src/service_discovery.cpp
@@ -1,18 +1,9 @@
 #include "plugin.h"
+#include "utils.h"
 
-#include <chrono>
 #include <cstring>
-#include <ctime>
 
 using json = nlohmann::json;
-
-static std::string nowTimestamp() {
-    auto now = std::chrono::system_clock::now();
-    std::time_t t = std::chrono::system_clock::to_time_t(now);
-    char buf[32];
-    std::strftime(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S", std::localtime(&t));
-    return buf;
-}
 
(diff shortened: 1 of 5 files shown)


src/service_discovery.cpp
```diff
@@ -1,9 +1,19 @@
...
 
 using json = nlohmann::json;
 
+static std::string nowTimestamp() {
+    auto now = std::chrono::system_clock::now();
+    std::time_t t = std::chrono::system_clock::to_time_t(now);
+    char buf[32];
+    std::strftime(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S", std::localtime(&t));
+    return buf;
+}
```
gmelodie: Can we extract this as well as other utilities to utils.cpp?

tinniaru3005: Sure! I'll move nowTimestamp() to utils and include it from service_discovery.cpp. Will update the PR.

gmelodie: You can use https://github.com/logos-co/logos-libp2p-module/pull/73 as a model

PR #77, opened 2026-07-01
feat: use nim-ffi cbindings

gmelodie:
## Summary

Adapts the module to nim-libp2p's nim-ffi cbindings (`chore/cbind/use-nim-ffi`), replacing the old flat `libp2p_*` C API. The generated header is now an async, typed, header-only binding (`LibP2PCtx*` + per-call `libp2p_ctx_<name>` wrappers, CBOR over the wire), so the whole wrapper layer was reworked around it.

## Changes

- **Async typed calls**: every op is `libp2p_ctx_<name>(ctx, [request], ReplyFn, ud)`; replies flow through per-response-type callbacks (`cbBool`, `cbBytes`, `cbPeerInfo`, `cbDial`, `cbRecords`, …) into a `SyncResult`. Success is `err_code == NIMFFI_RET_OK`.
- **Context/config**: `libp2p_ctx_t*` → `LibP2PCtx*` (delivered via the ctor callback); `libp2p_config_t` → `Libp2pConfig` with `NimFfiStr`/`NimFfiBytes`/`LibP2PSeq_*` views into member-held backing storage.
- **Streams**: the Nim side now owns streams and returns `uint64` ids — removed the local stream-pointer map and its locking; the wrapper forwards ids directly.
- **Events**: one `on_incoming_stream` and one `on_pubsub_message` listener registered at context creation, routing to `protocolStream` / `gossipsubMessage` events and the topic queue (guarded so a non-UTF-8 payload can't throw across the FFI boundary). Removed the old per-subscription/per-protocol callback plumbing and the dead `setEventCallback`.
- **Constants**: added `src/libp2p_consts.h` mirroring the `Direction_*` / `LIBP2P_TRANSPORT_*` / `LIBP2P_PK_*` / muxer ordinals the generated header no longer emits.
- **encode/decode XPR**: `createXpr` retained; `decodeXpr` wired to the `libp2p_ctx_decode_xpr` binding.
- **peerInfo naming**: calls `libp2p_ctx_peer_info` — nim-ffi snake-cases the `libp2pPeerInfo` proc to `peer_info`, so the old `libp2p_ctx_peerinfo` spelling no longer exists.

## nim-libp2p dependency

`flake.lock` pins `chore/cbind/use-nim-ffi` at the rev that ships the nim-ffi headers installed flat under `$out/include` (so the module builder's `cp include/*.h` staging resolves `lib/libp2p.h`) and the `libp2pDecodeXpr` proc — both from vacp2p/nim-libp2p#2681.

`tinycbor` — the CBOR codec the generated header's inline encoders/decoders call — is sourced from nixpkgs via `metadata.json` `nix.packages.runtime` and linked through `CMakeLists.txt` / `tests/CMakeLists.txt`.

## Testing

`nix flake check` passes locally against the pinned dependency (flake checks, asan/tsan sanitizers, and the openmetrics + standalone e2e apps).

The three blockers noted in earlier revisions of this description — the missing `libp2pDecodeXpr` proc, the nested-header install layout, and tinycbor not being on the consumer side — are all resolved; the `libp2p_ctx_peerinfo` → `libp2p_ctx_peer_info` rename is the final adaptation.


src/plugin.cpp
```diff
@@ -203,55 +181,14 @@ StdLogosResult Libp2pModuleImpl::createNode(const std::string& config) {
...
-        for (auto& t : pool) {
-            if (t.joinable()) t.join();
-        }
-    }
-
     if (ctx) {
-        destroyHandle(ctx);
+        // Synchronous: runs the Nim destructor (which drops the node and its
+        // streams) and frees the C-side context wrapper and listener boxes.
+        libp2p_ctx_destroy(ctx);
```
richard-ramos: Something to take into account:

libp2p_destroy returns an int with the status code, which indicates the code might be succesful or might fail,
. 
However, the code here: https://github.com/logos-messaging/nim-ffi/blob/7ef58f4bf5cf683cfdf2abbdb3d7688727763d65/ffi/codegen/c.nim#L558  does ignore that return code, thus there might be a failure here that's ignored.

Probably makes sense to fix that in nim-ffi?

src/plugin.cpp
```diff
@@ -74,113 +60,105 @@ Libp2pModuleImpl::Libp2pModuleImpl(const Libp2pModuleOptions& options)
...
-            return {false, {}, m_initError};
-        }
-        m_privKey.assign(keyResult.buffer.begin(), keyResult.buffer.end());
-    }
-
-    m_libp2pConfig.priv_key.data = m_privKey.data();
-    m_libp2pConfig.priv_key.dataLen = static_cast<int>(m_privKey.size());
+    // An empty privKey leaves the seq nil, so the Nim side generates a fresh
+    // identity; a supplied key gives a stable peer id across restarts.
+    m_libp2pConfig.privKey =
```
richard-ramos: should we pass the `m_libp2pConfig.m_keyType` too?

src/plugin.cpp
```diff
@@ -74,113 +60,105 @@ Libp2pModuleImpl::Libp2pModuleImpl(const Libp2pModuleOptions& options)
...
+        }
+        m_initError = "libp2p_ctx_create failed to submit (ret=" + std::to_string(ret) + ")";
+        fprintf(stderr, "libp2p_module: %s\n", m_initError.c_str());
         return {false, {}, m_initError};
     }
 
-    if (!candidate) {
-        m_initError = "libp2p_new returned null context";
-        fprintf(stderr, "libp2p_new returned null context\n");
+    auto r = awaitResult(f, kNewContextTimeoutMs);
```
richard-ramos, line 160: Curious about this `awaitResult`. What would happen if for ex, createContext for any reason takes more than kNEwContextTimeoutMs. Would it be garbage collected somehow?

src/plugin.cpp
```diff
@@ -287,20 +224,36 @@ StdLogosResult Libp2pModuleImpl::status() {
...
-        delete p;
-        return {false, "Failed to generate private key (ret=" + std::to_string(ret) + ")", {}, nullptr};
+    int ret = libp2p_ctx_new_private_key(ctx, &req, &Libp2pModuleImpl::cbBytes, p);
+    if (ret != 0) {
+        if (f.wait_for(std::chrono::seconds(0)) != std::future_status::ready) {
+            delete p;
+        }
+        SyncResult r;
+        r.message = "Failed to generate private key (ret=" + std::to_string(ret) + ")";
+        return r;
```
richard-ramos: Why is a context required to create a private key at all? Having memory allocated for a node shouldnt be required for this function.

Wouldnt this also have the side effect that if you create a private key first, then if you call createNode() it will fail with the "node already created" error?

src/stream.cpp
```diff
@@ -1,108 +1,59 @@
...
+    });
+}
+
+StdLogosResult Libp2pModuleImpl::streamCloseWithEOF(uint64_t streamId) {
+    return callSync("Failed to close stream with EOF", [&](SyncPromise* p) {
+        return libp2p_ctx_stream_close_with_eof(ctx, streamId, &Libp2pModuleImpl::cbBool, p);
+    });
+}
+
+StdLogosResult Libp2pModuleImpl::streamRelease(uint64_t streamId) {
```
richard-ramos, line 73: Previously there was a lock that helped avoid releasing the lock while the stream functions were being used, but now it got removed.  This is annoying because it means that Release will no longer wait until every active function is done with it. The nim side does not seem to have a way to handle this either...

gmelodie: Yes, but iiuc the issue would be a possible use-after-free? Now we've got a reference-counted `StreamRegistry`, the C++ side doesn't handle stream pointers at all.

gmelodie: So yeah now a release won't wait until every active function is done (waiting functions will error with `LPStreamError`, but there isn't a risk of a UAF at least for this case.

richard-ramos (approved):
LGTM

PR #80, opened 2026-07-21
feat(tutorial) add support for runnable tutorials with CI and docs generation 

vladopajic:
part 1 of https://github.com/logos-co/logos-libp2p-module/pull/78

### DIFF
--- a/.github/workflows/tutorial.yml
+++ b/.github/workflows/tutorial.yml
@@ -83,27 +83,3 @@ jobs:
             exit 1
           fi
           echo "✅ Generated markdown matches committed version."
-
-  upload-docs:
-    name: Upload generated documentation
-    runs-on: ubuntu-latest
-    timeout-minutes: 10
-    needs: [compile-check, markdown-check]
-    if: github.ref == 'refs/heads/master' && success()
-
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Generate markdown
-        run: |
-          chmod +x ./tutorial/generate_markdown.sh
-          ./tutorial/generate_markdown.sh
-
-      - name: Upload docs artifact
-        uses: actions/upload-artifact@v4
-        with:
-          name: tutorial-docs
-          path: tutorial/docs/
-          retention-days: 30
-          if-no-files-found: error
--- a/.github/workflows/tutorial.yml
+++ b/.github/workflows/tutorial.yml
@@ -1,0 +1,109 @@
+name: Tutorial CI
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - 'tutorial/**'
+      - 'src/**'
+      - 'CMakeLists.txt'
+      - 'tutorial/CMakeLists.txt'
+  pull_request:
+    paths:
+      - 'tutorial/**'
+      - 'src/**'
+      - 'CMakeLists.txt'
+      - 'tutorial/CMakeLists.txt'
+  workflow_dispatch:
+
+concurrency:
+  group: tutorial-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  NIX_CONFIG: |
+    experimental-features = nix-command flakes
+
+jobs:
+  compile-check:
+    name: Compile check (${{ matrix.name }})
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 30
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Linux (x86_64-linux)
+            runner: ubuntu-latest
+            system: x86_64-linux
+
+          - name: macOS (aarch64-darwin)
+            runner: macos-latest
+            system: aarch64-darwin
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v12
+
+      - name: Setup Nix Cache
+        uses: DeterminateSystems/magic-nix-cache-action@v5
+
+      - name: Build tutorials inside Nix dev shell
+        run: |
+          nix develop --command ./tutorial/build_tutorials.sh
+
+  markdown-check:
+    name: Markdown freshness check
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Generate markdown from tutorial sources
+        run: |
+          chmod +x ./tutorial/generate_markdown.sh
+          rm -rf ./tutorial/docs
+          ./tutorial/generate_markdown.sh
+
+      - name: Check git diff (no changes expected)
+        run: |
+          if ! git diff --quiet -- ./tutorial/docs/; then
+            echo "❌ Generated markdown does not match committed version."
+            echo ""
+            echo "Differences found:"
+            git diff -- ./tutorial/docs/
+            echo ""
+            echo "Run './tutorial/generate_markdown.sh' locally and commit the updated docs/"
+            exit 1
+          fi
+          echo "✅ Generated markdown matches committed version."
+
+  upload-docs:
+    name: Upload generated documentation
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    needs: [compile-check, markdown-check]
+    if: github.ref == 'refs/heads/master' && success()
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Generate markdown
+        run: |
+          chmod +x ./tutorial/generate_markdown.sh
+          ./tutorial/generate_markdown.sh
+
+      - name: Upload docs artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: tutorial-docs
+          path: tutorial/docs/
+          retention-days: 30
+          if-no-files-found: error
--- a/src/utils.h
+++ b/src/utils.h
@@ -9,7 +9,7 @@
 #include <nlohmann/json.hpp>
 
 extern "C" {
-#include "lib/libp2p.h"
+#include <libp2p.h>
 }
 
 // Borrows each string's c_str() into the const char* array the cbinding layer
--- a/tutorial/README.md
+++ b/tutorial/README.md
@@ -22,7 +22,6 @@ nix --extra-experimental-features 'nix-command flakes' develop --command ./tutor
 ```
 
 **Build prerequisites**
-- `logos-libp2p-module` built with CMake (see root `README.md` for build instructions)
 - Nix development shell (`nix develop`) or equivalent dependencies
 
 No need to rebuild between runs unless the source code changes.
--- a/tutorial/README.md
+++ b/tutorial/README.md
@@ -8,27 +8,37 @@ Step-by-step tutorials for the `logos-libp2p-module` — from basic node lifecyc
 The tutorials are compiled alongside the module. Build them **one time**, then run any tutorial immediately:
 
 ```bash
+# Build once
 nix develop
 ./tutorial/build_tutorials.sh
+
+# or use following command if you are missing experimental Nix features:
+nix --extra-experimental-features 'nix-command flakes' develop --command ./tutorial/build_tutorials.sh
+
 
 # Run any tutorial:
 ./build/tutorial/tutorial_1_node_lifecycle
 ./build/tutorial/tutorial_3_connecting_peers
-# ...
 ```
+
+**Build prerequisites**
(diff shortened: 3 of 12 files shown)


src/utils.h
```diff
@@ -9,7 +9,7 @@
 #include <nlohmann/json.hpp>
 
 extern "C" {
-#include "lib/libp2p.h"
+#include <libp2p.h>
```
gmelodie, line 12: nice!

tutorial/build_tutorials.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+#
+# build_tutorials.sh
+#
+# Builds all tutorial executables using the project's CMake build system.
+# Should be invoked within the Nix development shell (nix develop) so that
+# LOGOS_MODULE_BUILDER_ROOT, libp2p.so, and other dependencies are available.
```
gmelodie, line 7: Maybe this could be a check with an error message.

tutorial/build_tutorials.sh
```diff
@@ -0,0 +1,72 @@
...
+#
+# Usage:
+#   nix --extra-experimental-features 'nix-command flakes' develop --command ./tutorial/build_tutorials.sh
+#   # or from inside the dev shell:
+#   ./tutorial/build_tutorials.sh
```
gmelodie: maybe we could have a `--help/-h` tag with this text (that also just gets called when arguments are wrong)

tutorial/build_tutorials.sh
```diff
@@ -0,0 +1,72 @@
...
+#
+# The script cleans only the tutorial targets so existing test/example builds
+# remain intact.
+#
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+PROJECT_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
```
gmelodie: a bit of hardening: silence cd and guard against paths starting with a dash
```suggestion
SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
PROJECT_DIR="$(cd -- "${SCRIPT_DIR}/.." &>/dev/null && pwd)"
```

tutorial/build_tutorials.sh
```diff
@@ -0,0 +1,72 @@
...
+
+cd "${PROJECT_DIR}"
+
+# Build directory (same as the project's standard cmake build)
+BUILD_DIR="${PROJECT_DIR}/build"
+
+# Ensure cmake is configured. A previous failed configure can leave a
+# CMakeCache.txt without a generated build system.
+if [ ! -f "${BUILD_DIR}/CMakeCache.txt" ] || [ ! -f "${BUILD_DIR}/Makefile" ]; then
```
gmelodie: I wonder if this is enough... I mean can the cache be in a half-finished state? if this does not take long, should we just get rid of the files and always rebuild?

(nit?)

tutorial/build_tutorials.sh
```diff
@@ -0,0 +1,72 @@
...
+if [ "${#tutorial_targets[@]}" -eq 0 ]; then
+    echo "No tutorial sources found in ${SCRIPT_DIR}" >&2
+    exit 1
+fi
+
+cmake --build "${BUILD_DIR}" --target "${tutorial_targets[@]}" \
+    -j "$(nproc 2>/dev/null || echo 4)"
+
+echo ""
+echo "✅ All tutorials compiled successfully."
```
gmelodie, line 73:
```suggestion
echo "All tutorials compiled successfully."
```

tutorial/README.md
```diff
@@ -0,0 +1,45 @@
+# logos-libp2p-module Tutorials
```
gmelodie: nit
```suggestion
# Tutorials
```

tutorial/README.md
```diff
@@ -0,0 +1,45 @@
...
+nix develop
+./tutorial/build_tutorials.sh
+
+# Run any tutorial:
+./build/tutorial/tutorial_1_node_lifecycle
+./build/tutorial/tutorial_3_connecting_peers
+# ...
+```
+
+No need to rebuild between runs unless the source code changes.
```
gmelodie, line 27:
```suggestion
## Running

Build all tutorials:

```bash
nix develop
./tutorial/build_tutorials.sh

# Run a tutorial:
./build/tutorial/tutorial_1_node_lifecycle
./build/tutorial/tutorial_3_connecting_peers
# ...
```

```

1. makes me wonder if we want to build a single tutorial? maybe not, the nix deps is the part that takes a while but still, we could check

2. nit: thinking if `build_tutorials` could be `run_tutorials` instead and build if needed, but probably more work than necessary

File: tutorial/README.md [hunk 9]
```diff
@@ -0,0 +1,45 @@
...
+
+## Workflow
+
+1. **Read** a tutorial page (start with [Tutorial 1](docs/tutorial_1_node_lifecycle.md))
+2. **Run** the matching binary to see it in action
+3. **Read the source** in the `.cpp` file for the full code
```
gmelodie:
```suggestion
```

tutorial/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Prerequisites
+
+- `logos-libp2p-module` built with CMake (see root `README.md` for build instructions)
+- Nix development shell (`nix develop`) or equivalent dependencies
```
gmelodie: This should probably be before build/run instructions

tutorial/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+The `docs/` folder is generated from `///` comments in the `.cpp` sources:
+
+```bash
+./tutorial/generate_markdown.sh
+```
+
+<div align="center">
+<br>
+<a href="docs/tutorial_1_node_lifecycle.md"><b>▶ START HERE — Tutorial 1: Node Lifecycle</b></a>
```
gmelodie:
```suggestion
<a href="docs/tutorial_1_node_lifecycle.md"><b>START HERE — Tutorial 1: Node Lifecycle</b></a>
```

tutorial/tutorial_1_node_lifecycle.cpp
```diff
@@ -0,0 +1,106 @@
...
+///
+/// Make sure you can build the `logos-libp2p-module` project. See the
+/// project's `README.md` for build instructions using Nix or CMake.
+///
+/// ## What is libp2p?
+///
+/// [libp2p](https://libp2p.io/) is a modular networking stack for building
+/// peer-to-peer applications. It provides transport-agnostic connectivity,
+/// peer identity, stream multiplexing, secure channels, and content routing
+/// — everything you need to build a decentralized network.
```
gmelodie, line 18:
```suggestion
/// peer-to-peer applications. It provides building blocks to create decentralized networks, such as transport-agnostic connectivity,
/// peer identity, stream multiplexing, secure channels, and content routing,
```

tutorial/tutorial_1_node_lifecycle.cpp
```diff
@@ -0,0 +1,106 @@
...
+///
+/// The `logos-libp2p-module` wraps libp2p's C bindings into a C++ class
```
gmelodie:
```suggestion
/// The `logos-libp2p-module` wraps nim-libp2p's C bindings into a C++ class
```

tutorial/generate_markdown.sh
```diff
@@ -0,0 +1,240 @@
...
+# Usage:
+#   ./generate_markdown.sh                       # generate all .md files
+#   ./generate_markdown.sh tutorial_1.cpp         # generate a single file
+#
+# Output: tutorial/docs/tutorial_<name>.md for each input file
+#
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
```
gmelodie: same as above

.github/workflows/tutorial.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Generate markdown from tutorial sources
+        run: |
+          chmod +x ./tutorial/generate_markdown.sh
+          rm -rf ./tutorial/docs
+          ./tutorial/generate_markdown.sh
+
+      - name: Check git diff (no changes expected)
+        run: |
+          if ! git diff --quiet -- ./tutorial/docs/; then
+            echo "❌ Generated markdown does not match committed version."
```
gmelodie, line 77:
```suggestion
            echo "Generated markdown does not match committed version."
```

.github/workflows/tutorial.yml
```diff
@@ -0,0 +1,85 @@
...
+            echo ""
+            echo "Differences found:"
+            git diff -- ./tutorial/docs/
+            echo ""
+            echo "Run './tutorial/generate_markdown.sh' locally and commit the updated docs/"
+            exit 1
+          fi
+          echo "✅ Generated markdown matches committed version."
```
gmelodie, line 85:
```suggestion
          echo "Generated markdown matches committed version."
```

gmelodie (commented):
Also a few overall comments that I found useful from claude:
- Replace `echo "$x" | grep -q 'pat'` with `[[ $x == pat* ]]` or `case` — kills ~2 forks per line.
- Replace `echo "$x" | sed 's/^foo//'` with `${x#foo}` — parameter expansion, no subprocess.
- Never `echo "$var"` for arbitrary data; use `printf '%s\n' "$var"` (leading `-n/-e` and backslashes get eaten).
- `$0` → `${BASH_SOURCE[0]}` so the path still resolves when sourced.
- Script is GNU-only (`\+`, `\b`, `\u` in `sed`) despite env bash — either drop them or state the dependency.
- Prefer `[[ ]]` over `[ ]` throughout in a Bash-only script — no word-splitting surprises, supports `==`/`=~`.

.github/workflows/tutorial.yml
vladopajic: i prefere it with fancy icon for status. is makes a differnece when using script. since i used it offten  while developing, i liked it to see status right away.

sometimes people put too much icons all over the place, this is just usability.

i'll leave it for richard to vote.

tutorial/build_tutorials.sh
vladopajic: i would prefer simplicity with less code here here. adding additional commands to this shell is unnecessary as usage is fixed to one use-case in this repo.

tutorial/build_tutorials.sh
vladopajic: this script was ai generated. i wouldn't send too much time perfecting it, unless there is tangible concern, since it works, exposure and usage is limited.

tutorial/README.md
vladopajic: removed.

tutorial/README.md
vladopajic: moved inside build/run instructions

tutorial/README.md
vladopajic: this section is needed as i really want to suggest workflow. i want user to read it to read it to have mental model how to go through tutorial. they don't really have to follow it, but if they read this they will have high level concept of how tutorials are structured.

also added more text.

tutorial/README.md
vladopajic: 1. no difference for building single tutorial and building all - hence building all
2. as commented "probably more work than necessary"

tutorial/build_tutorials.sh
vladopajic: added nix check below

tutorial/build_tutorials.sh
vladopajic: also nix check https://github.com/logos-co/logos-libp2p-module/pull/80/commits/2176d49e25b81378ccfbe8e6699bb55a053c0522 solves this in different way

tutorial/README.md
```diff
@@ -0,0 +1,55 @@
...
+nix --extra-experimental-features 'nix-command flakes' develop --command ./tutorial/build_tutorials.sh
+
+
+# Run any tutorial:
+./build/tutorial/tutorial_1_node_lifecycle
+./build/tutorial/tutorial_3_connecting_peers
+```
+
+**Build prerequisites**
+- `logos-libp2p-module` built with CMake (see root `README.md` for build instructions)
```
richard-ramos: I find this a bit confusing because this built prerequisite references CMake, yet on the build instructions `nix build` is used (CMake should already be provided by nix).

richard-ramos (approved):
Image

tutorial/README.md
vladopajic: yeah, i see wym. this item could be left out

.github/workflows/tutorial.yml
richard-ramos: https://www.youtube.com/watch?v=ussCHoQttyQ

PR #81, opened 2026-07-22
docs(tutorial): add tutorials 2, 3 and 4

vladopajic:
part #2 of #78

adds only tutorial source for 2, 3 and 4

### DIFF
--- a/tutorial/docs/tutorial_2_custom_config.md
+++ b/tutorial/docs/tutorial_2_custom_config.md
@@ -1,0 +1,180 @@
+# Tutorial 2: Custom Node Configuration
+
+In the [previous tutorial](tutorial_1_node_lifecycle.md), we created a
+node with default settings that binds to a random port. In this tutorial,
+we'll learn how to configure the node explicitly — setting a fixed port,
+changing transports, and passing options via both C++ structs and JSON.
+
+## Understanding Libp2pModuleOptions
+
+The `Libp2pModuleImpl` constructor accepts a `Libp2pModuleOptions` struct
+that lets you control every aspect of the node. Here are the most common
+configuration fields:
+
+| Field | Type | Default | Description |
+|-------|------|---------|-------------|
+| `addrs` | `vector<string>` | `["/ip4/127.0.0.1/tcp/0"]` | Listen addresses |
+| `transport` | `int` | `LIBP2P_TRANSPORT_TCP` | Transport protocol |
+| `maxConnections` | `int` | `50` | Max total connections |
+| `maxConnsPerPeer` | `int` | `1` | Max connections per peer |
+| `mountGossipsub` | `bool` | `true` | Enable GossipSub |
+| `mountKad` | `bool` | `true` | Enable Kademlia DHT |
+| `mountServiceDiscovery` | `bool` | `true` | Enable service discovery |
+
+## Binding to a fixed port
+
+By default nodes bind to port 0 (random). To listen on a specific port,
+provide a custom address in the `addrs` field.
+```cpp
+#include <cstdio>
+#include <string>
+#include "plugin.h"
+
+int main()
+{
+```
+
+## Method 1: Configure via C++ struct
+
+The most explicit way to configure a node is by constructing
+`Libp2pModuleOptions` directly and passing it to the constructor.
+```cpp
+    Libp2pModuleOptions options;
+
+    // Listen on TCP port 9090 on all interfaces
+    options.addrs = {"/ip4/0.0.0.0/tcp/9090"};
+
+    // Limit to 10 total connections
+    options.maxConnections = 10;
+
+    // Allow up to 2 connections from the same peer
+    options.maxConnsPerPeer = 2;
+
+    // We don't need GossipSub or Kademlia for this basic example
+    options.mountGossipsub = false;
+    options.mountKad = false;
+    options.mountServiceDiscovery = false;
+
+    printf("Creating node with custom configuration...\n");
+    printf("  Listen address: %s\n", options.addrs[0].c_str());
+    printf("  Max connections: %d\n", options.maxConnections);
+    printf("  Max per peer: %d\n", options.maxConnsPerPeer);
+
+    Libp2pModuleImpl node(options);
+
+    if (!node.start().success) {
+        fprintf(stderr, "Failed to start node\n");
+        return 1;
+    }
+
+```
+
+Verify that the node is actually listening on the configured port:
+```cpp
+    auto info = node.peerInfo();
+    if (info.success) {
+        printf("\nNode is live!\n");
+        printf("  Peer ID: %s\n",
+               info.value["peerId"].get<std::string>().c_str());
+        for (const auto& addr : info.value["addrs"]) {
+            printf("  Listening on: %s\n",
+                   addr.get<std::string>().c_str());
+        }
+    }
+
+    node.stop();
+
+```
+
+## Method 2: Configure via JSON
+
+When integrating with larger systems (e.g. `logoscore`), you'll
+often receive configuration as JSON. The same options can be
+supplied via `Libp2pModuleOptions::fromJson()`.
+```cpp
+    printf("\n--- Method 2: JSON configuration ---\n");
+
+    std::string jsonConfig = R"({
+        "addrs": ["/ip4/127.0.0.1/tcp/9091"],
+        "maxConnections": 20,
+        "mountGossipsub": false,
+        "mountKad": false,
+        "mountServiceDiscovery": false
+    })";
+
+    bool parseOk;
+    std::string parseErr;
+    auto jsonOpts = Libp2pModuleOptions::fromJson(jsonConfig, parseOk, &parseErr);
+
+    if (!parseOk) {
+        fprintf(stderr, "Failed to parse JSON config: %s\n",
+                parseErr.c_str());
+        return 1;
+    }
+
+    printf("Parsed JSON config:\n");
+    printf("  Listen address: %s\n", jsonOpts.addrs[0].c_str());
+    printf("  Max connections: %d\n", jsonOpts.maxConnections);
+
+    Libp2pModuleImpl nodeFromJson(jsonOpts);
+
+    if (!nodeFromJson.start().success) {
+        fprintf(stderr, "Failed to start JSON-configured node\n");
+        return 1;
+    }
+
+    auto info2 = nodeFromJson.peerInfo();
+    if (info2.success) {
+        printf("\nJSON-configured node is live!\n");
+        printf("  Peer ID: %s\n",
+               info2.value["peerId"].get<std::string>().c_str());
+        for (const auto& addr : info2.value["addrs"]) {
+            printf("  Listening on: %s\n",
+                   addr.get<std::string>().c_str());
+        }
+    }
+
+    nodeFromJson.stop();
+
+    return 0;
+}
+
+```
+
+## Exercise: Try different configurations
+
+Experiment with these variations:
+
+```cpp
+// QUIC transport instead of TCP
+options.transport = LIBP2P_TRANSPORT_QUIC;
+options.addrs = {"/ip4/127.0.0.1/udp/9092/quic-v1"};
+
+// Listen on multiple addresses
+options.addrs = {
+    "/ip4/0.0.0.0/tcp/9093",
+    "/ip4/0.0.0.0/tcp/9094"
+};
+```
+
+## Summary
+
+In this tutorial you learned:
+  - How to configure a node with a fixed port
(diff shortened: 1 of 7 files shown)


tutorial/docs/tutorial_2_custom_config.md
```diff
@@ -0,0 +1,180 @@
...
+By default nodes bind to port 0 (random). To listen on a specific port,
+provide a custom address in the `addrs` field.
+```cpp
+#include <cstdio>
+#include <string>
+#include "plugin.h"
+
+int main()
+{
+```
```
gmelodie, line 35: ?

tutorial/docs/tutorial_2_custom_config.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Method 1: Configure via C++ struct
+
+The most explicit way to configure a node is by constructing
+`Libp2pModuleOptions` directly and passing it to the constructor.
```
gmelodie, line 40: maybe we could link the next section here already like


```suggestion
The most explicit way to configure a node is by constructing
`Libp2pModuleOptions` directly and passing it to the constructor, but it can also be configured [via json](...)
```

tutorial/docs/tutorial_2_custom_config.md
```diff
@@ -0,0 +1,180 @@
...
+    // Listen on TCP port 9090 on all interfaces
+    options.addrs = {"/ip4/0.0.0.0/tcp/9090"};
+
+    // Limit to 10 total connections
+    options.maxConnections = 10;
+
+    // Allow up to 2 connections from the same peer
+    options.maxConnsPerPeer = 2;
+
+    // We don't need GossipSub or Kademlia for this basic example
```
gmelodie, line 53:
```suggestion
    // We don't need GossipSub or Kademlia for this example
```

tutorial/docs/tutorial_2_custom_config.md
```diff
@@ -0,0 +1,180 @@
+# Tutorial 2: Custom Node Configuration
```
gmelodie, line 1: I think we should link the entire code file here at the top of the tutorial too

gmelodie: This is for all tutorials

tutorial/docs/tutorial_2_custom_config.md
```diff
@@ -0,0 +1,180 @@
...
+
+```
+
+Verify that the node is actually listening on the configured port:
+```cpp
+    auto info = node.peerInfo();
+    if (info.success) {
+        printf("\nNode is live!\n");
+        printf("  Peer ID: %s\n",
+               info.value["peerId"].get<std::string>().c_str());
```
gmelodie, line 78: This is really ugly, maybe we should add some helpers of, if there's a more elegant solution, do that instead

tutorial/docs/tutorial_3_connecting_peers.md
```diff
@@ -0,0 +1,226 @@
...
+#include <cstdint>
+#include <string>
+#include <vector>
+#include "plugin.h"
+
+int main()
+{
+    printf("=== Tutorial 3: Connecting Peers ===\n\n");
+
+```
```
gmelodie, line 44: ?

tutorial/docs/tutorial_4_custom_protocol.md
```diff
@@ -0,0 +1,261 @@
...
+for request-response protocols.
+
+```cpp
+const std::string kEchoProtocol = "/examples/echo/1.0.0";
+
+int main()
+{
+    printf("=== Tutorial 4: Custom Protocol Handlers ===\n\n");
+
+```
```
gmelodie, line 58: ?

tutorial/docs/tutorial_4_custom_protocol.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Step 9: Clean up
+
+The client (dialing side) closes its stream with `streamCloseWithEOF()`
+to signal that it's done writing. Both sides release their stream
+resources with `streamRelease()`.
+```cpp
+    nodeB.streamCloseWithEOF(clientStreamId);
+    nodeB.streamRelease(clientStreamId);
+    nodeA.streamRelease(serverStreamId);
```
gmelodie: Do we have to do this?

tutorial/docs/tutorial_2_custom_config.md
vladopajic: please explain

tutorial/docs/tutorial_3_connecting_peers.md
vladopajic: please explain

tutorial/docs/tutorial_4_custom_protocol.md
vladopajic: please explain

tutorial/docs/tutorial_2_custom_config.md
vladopajic: i don't think tutorial should use helpers at all. tutorial should be based on raw library usage.

tutorial/docs/tutorial_2_custom_config.md
vladopajic: not sure about this. i had this in mind this but decided not to complicate.

the way tutorials are going to be view is likely going to be via browser (?). and if they want to run code, they will have to clone repo and use ide/console. 

second thing is that final step might remove all `.md` files and host it as github pages.

tutorial/tutorial_4_custom_protocol.cpp
```diff
@@ -0,0 +1,214 @@
...
+    printf("Echo verified successfully!\n");
+
+/// ## Step 9: Clean up
+///
+/// The client (dialing side) closes its stream with `streamCloseWithEOF()`
+/// to signal that it's done writing. Both sides release their stream
+/// resources with `streamRelease()`.
+    nodeB.streamCloseWithEOF(clientStreamId);
+    nodeB.streamRelease(clientStreamId);
+    nodeA.streamRelease(serverStreamId);
```
richard-ramos: Hm... there might be a deadlock here because streamCloseWithEOF waits until the server sides closes. (there's a comment in this .md file that says `Unlike the dialing side, the server side does **not** call streamClose() — the peer that initiated the stream is responsible for closing it.`

So... maybe this should be changed so it matches the order seen in the test https://github.com/logos-co/logos-libp2p-module/blob/master/tests/custom_handlers.cpp#L100-L104

```
nodeA.streamRelease(serverStreamId);
nodeB.streamCloseWithEOF(clientStreamId);
nodeB.streamRelease(clientStreamId);
```

tutorial/docs/tutorial_3_connecting_peers.md
```diff
@@ -0,0 +1,226 @@
...
+        return 1;
+    }
+    printf("Connected!\n");
+
+```
+
+## Step 4: List connected peers
+
+We can verify the connection by listing connected peers on each node.
+The direction parameter `0` means "all connected peers".
```
richard-ramos: 0 means peers connected to you (Direction_In), not all connected peers. Altho this raises some interesting points:

1 - Maybe we should in the bindings have a way to get all peers regardless of the direction.
2 - Previously we did have some constants in the bindings in libp2p.h for enums (so we dont have to use magic numbers). Is this possible to do with nim-ffi, @gmelodie 
3 - Also, in line 129, since here B is connection to A, the direction should be 1  (Direction_Out)

tutorial/docs/tutorial_4_custom_protocol.md
```diff
@@ -0,0 +1,261 @@
...
+
+## How Custom Protocols Work
+
+A protocol in libp2p is identified by a **protocol ID string** — a
+`/`-separated path like `/myapp/chat/1.0.0`. When a remote peer dials
+this protocol ID, your node receives a new stream.
+
+To handle incoming streams you:
+  1. Call `mountProtocol()` to register a protocol ID
+  2. Set an `emitEvent` callback that listens for `"protocolStream"` events
```
richard-ramos, line 18: 2 should happen before 1, otherwise mountProtocol will fail
https://github.com/logos-co/logos-libp2p-module/blob/5a777c1622f2cef34ec4a371427581d8945c5777/src/custom_handlers.cpp#L39

tutorial/docs/tutorial_4_custom_protocol.md
```diff
@@ -0,0 +1,261 @@
...
+  3. Read from and write to the stream in the event handler
```
richard-ramos: I am not sure about this line. In the event handler we should capture the streamId in the callback, but, the reading/writing should be done later because we gotta try not to do blocking operations inside event callbacks.

tutorial/docs/tutorial_2_custom_config.md
gmelodie: Apologies! I just didn´t understand what was this supposed to show, as it doesn't seem to be related to the text above it. (same for other question marks)

tutorial/docs/tutorial_2_custom_config.md
gmelodie: ah i see. okay!

tutorial/docs/tutorial_2_custom_config.md
gmelodie: oh i meant, add helpers for all users, not just tutorial. if anything, the tutorial revealed places where usage could be improved. i didn't mean to say that this should be done in this PR specifically, although it could, but if you agree we can create an issue

tutorial/docs/tutorial_3_connecting_peers.md
gmelodie: >  2 - Previously we did have some constants in the bindings in libp2p.h for enums (so we dont have to use magic numbers). Is this possible to do with nim-ffi, @gmelodie

Nope, pushing a PR with this too

> 3 - Since node-b is the one initiating the connection, probably we should use

Cliff hanger?

gmelodie (approved):
I think my remaining comments are nit-ish, so approving. thanks!

tutorial/docs/tutorial_3_connecting_peers.md
richard-ramos: Fixed, lol

gmelodie: 1. https://github.com/vacp2p/nim-libp2p/pull/2878
2. https://github.com/logos-messaging/nim-ffi/pull/140

tutorial/docs/tutorial_2_custom_config.md
vladopajic: all methods return `StdLogosResult` which holds untyped json value, so getting parts from results always has undergo type casting. 

the only way around it is to refactor library methods to return types for each response. all those types could inherit `StdLogosResult` if this type of response is needed for some generic integration?

tutorial/docs/tutorial_3_connecting_peers.md
vladopajic: for tutorial, i have updated text here.

should this pr be blocked or enums to be merged, or we can do it later?

tutorial/docs/tutorial_4_custom_protocol.md
vladopajic: fixed

tutorial/docs/tutorial_3_connecting_peers.md
richard-ramos: Can be done later!

tutorial/docs/tutorial_4_custom_protocol.md
vladopajic: in order to make tutorial easy digestible (in terms of total code), i suggest we keep tutorial as is - do business logic in event handlers. currently goal of tutorial was meant to get beginners of the ground.

that said, it makes sense to add note to this point, and explain user implication of this code.

richard-ramos (approved):

tutorial/tutorial_4_custom_protocol.cpp
vladopajic: improved as suggested

tutorial/docs/tutorial_4_custom_protocol.md
vladopajic: yes, https://github.com/logos-co/logos-libp2p-module/pull/81#discussion_r3632434494

tutorial/docs/tutorial_3_connecting_peers.md
vladopajic: https://github.com/logos-co/logos-libp2p-module/issues/83

PR #82, opened 2026-07-22
feat: generic custom-protocol bridge for cross-module consumers

adklempner:
## Generic custom-protocol bridge for cross-module consumers

Adds content-agnostic, single-JSON-blob methods so a **consumer module** can drive
any length-prefixed request/response protocol over the module boundary — **without
adding protocol-specific code to `libp2p_module`**:

| Method | Shape |
|---|---|
| `protocolRequest` | `{peerId, multiaddrs[], proto, requestB64, timeoutMs?, maxSize?} -> {responseB64}` — connect + dial + writeLp + readLp + release (client one-shot) |
| `streamReadLpJson` | `{streamId, maxSize?, timeoutMs?} -> {dataB64}` |
| `streamWriteLpJson` | `{streamId, dataB64} -> {}` (base64-decoded in C++) |
| `streamCloseJson` / `streamReleaseJson` | `{streamId}` |
| `protocolAcceptStream` | `{proto, timeoutMs?} -> {streamId, proto}` — polling drain of inbound streams |

### Why

The typed `dial`/`stream*` methods take multi-string / `uint64` signatures the
universal codegen QtRO dispatch can't marshal across the module boundary, and
`streamWriteLp` writes **raw** bytes that don't survive a JSON string arg. These
wrappers carry **opaque base64** instead, so any binary protocol survives the trip.
Everything builds on the existing base primitives (`connectPeer`, `dial`,
`streamWriteLp`, `callSyncStreamWith`, `base64*`) — no new dependencies.

### Inbound delivery: polling, not push

`protocolAcceptStream` drains a per-proto queue that `protocolHandler` now fills,
mirroring the existing `gossipsubNextMessage` pattern. This is deliberate: the
`protocolStream` push-event doesn't cross the module boundary (the universal
codegen skips the `emitEvent` sink), so a consumer module has no way to receive
it. Polling is the same mechanism gossipsub already uses successfully.

### Behavior change (please note)

`mountProtocol` **no longer requires `emitEvent` to be set.** Previously it
returned an error when `emitEvent` was unset, to avoid leaking a stream no one
could read. That's now safe because every inbound stream is queued for
`protocolAcceptStream` regardless — and the `protocolStream` event is *still*
emitted when `emitEvent` is set, so in-process consumers (tests/examples) are
unaffected. The existing `custom_handlers_mount_requires_emit_event` test is
updated to assert the new contract (`..._mount_without_emit_event_ok`).

### Tests

`tests/custom_handlers.cpp` adds `protocol_bridge_request_accept_roundtrip`: two
nodes complete a full request→echo→response using **only** the new JSON-blob
methods (server polls `protocolAcceptStream` + `streamReadLpJson` /
`streamWriteLpJson`, client drives one `protocolRequest`), with no `emitEvent`
and no typed calls — the exact cross-module usage shape, exercised in-process.
`src/protocol_bridge.cpp` is added to the integration test build.

### Footprint

+~310 / −8: `src/protocol_bridge.cpp` (new), `src/custom_handlers.cpp`,
`src/plugin.h`, `CMakeLists.txt`, plus `tests/custom_handlers.cpp` /
`tests/CMakeLists.txt`. Validated end-to-end driving a real custom protocol (RLN
membership gifter) cross-module from a separate consumer module.

### DIFF
--- a/src/custom_handlers.cpp
+++ b/src/custom_handlers.cpp
@@ -2,6 +2,29 @@
 
 using json = nlohmann::json;
 
+static constexpr size_t kMaxInboundStreamsPerProto = 1024;
+
+bool Libp2pModuleImpl::enqueueInboundStream(const std::string& proto,
+                                            libp2p_stream_t* stream, uint64_t streamId) {
+    bool queued = false;
+    {
+        std::lock_guard<std::mutex> lock(m_inboundStreamMutex);
+        auto& q = m_inboundStreamQueues[proto];
+        if (q.size() < kMaxInboundStreamsPerProto) {
+            q.push_back(streamId);
+            queued = true;
+        }
+    }
+    if (!queued) {
+        removeStream(streamId);
+        libp2p_stream_release(ctx, stream,
+                              +[](int, const char*, size_t, void*) {}, nullptr);
+        return false;
+    }
+    m_inboundStreamCond.notify_all();
+    return true;
+}
+
 void Libp2pModuleImpl::protocolHandler(
     libp2p_ctx_t* /*ctx*/, libp2p_stream_t* stream,
     const char* proto, size_t protoLen, void* userData)
@@ -15,6 +38,7 @@ void Libp2pModuleImpl::protocolHandler(
         : handlerCtx->proto;
 
     uint64_t streamId = self->addStream(stream);
+    if (!self->enqueueInboundStream(protoStr, stream, streamId)) return;
 
     json j;
     j["streamId"] = streamId;
@@ -34,9 +58,6 @@ void Libp2pModuleImpl::mountCompleteCallback(int ret, const char* msg, size_t le
 StdLogosResult Libp2pModuleImpl::mountProtocol(const std::string& proto) {
     if (!ctx) return {false, {}, "No libp2p context"};
     if (proto.empty()) return {false, {}, "Protocol string is empty"};
-    // Without emitEvent, protocolHandler would register a stream that no caller
-    // could ever read, close, or release — leaking the stream.
-    if (!emitEvent) return {false, {}, "emitEvent must be set before mounting a protocol"};
     publishEmitEvent();
 
     auto handlerCtx = std::make_unique<ProtocolHandlerCtx>();
--- a/src/plugin.h
+++ b/src/plugin.h
@@ -6,6 +6,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstdlib>
+#include <deque>
 #include <functional>
 #include <future>
 #include <memory>
@@ -149,6 +150,13 @@ public:
     StdLogosResult streamCloseWithEOF(uint64_t streamId);
     StdLogosResult streamRelease(uint64_t streamId);
 
+    StdLogosResult protocolRequest(const std::string& argsJson);
+    StdLogosResult streamReadLpJson(const std::string& argsJson);
+    StdLogosResult streamWriteLpJson(const std::string& argsJson);
+    StdLogosResult streamCloseJson(const std::string& argsJson);
+    StdLogosResult streamReleaseJson(const std::string& argsJson);
+    StdLogosResult protocolAcceptStream(const std::string& argsJson);
+
     StdLogosResult gossipsubPublish(const std::string& topic, const std::string& data);
     StdLogosResult gossipsubSubscribe(const std::string& topic);
     StdLogosResult gossipsubUnsubscribe(const std::string& topic);
@@ -228,10 +236,17 @@ private:
     uint64_t addStream(libp2p_stream_t* stream);
     std::shared_ptr<StreamEntry> getStream(uint64_t id) const;
     std::shared_ptr<StreamEntry> removeStream(uint64_t id);
+    void removeInboundStream(uint64_t id);
+    bool enqueueInboundStream(const std::string& proto, libp2p_stream_t* stream,
+                              uint64_t streamId);
 
     std::mutex m_queueMutex;
     std::condition_variable m_queueCond;
     std::unordered_map<std::string, std::queue<std::string>> m_topicQueues;
+
+    std::mutex m_inboundStreamMutex;
+    std::condition_variable m_inboundStreamCond;
+    std::unordered_map<std::string, std::deque<uint64_t>> m_inboundStreamQueues;
 
     void applyOptions(const Libp2pModuleOptions& options);
     StdLogosResult createContext();
@@ -315,14 +330,16 @@ private:
     // whole sync-over-async call so a concurrent streamRelease can't free
     // entry->ptr mid-flight. `invoke(libp2p_stream_t*, SyncPromise*)`.
     template <class Invoke>
-    StdLogosResult callSyncStream(uint64_t streamId, const char* errPrefix, Invoke&& invoke) {
+    StdLogosResult callSyncStream(uint64_t streamId, const char* errPrefix, Invoke&& invoke,
+                                  int awaitMs = kDefaultOpTimeoutMs) {
         return callSyncStreamWith(streamId, errPrefix, std::forward<Invoke>(invoke),
-            [](const SyncResult&) -> StdLogosResult { return {true, {}, ""}; });
+            [](const SyncResult&) -> StdLogosResult { return {true, {}, ""}; }, awaitMs);
     }
 
     template <class Invoke, class Transform>
     StdLogosResult callSyncStreamWith(uint64_t streamId, const char* errPrefix,
-                                       Invoke&& invoke, Transform&& transform) {
+                                       Invoke&& invoke, Transform&& transform,
+                                       int awaitMs = kDefaultOpTimeoutMs) {
         if (!ctx || streamId == 0) return {false, {}, "Invalid stream"};
         auto entry = getStream(streamId);
         if (!entry) return {false, {}, "Stream not found"};
@@ -330,7 +347,7 @@ private:
         if (entry->released) return {false, {}, "Stream released"};
         return callSyncWith(errPrefix,
             [&](SyncPromise* p) { return invoke(entry->ptr, p); },
-            std::forward<Transform>(transform));
+            std::forward<Transform>(transform), awaitMs);
     }
 
     // Subscribe contexts live for the subscription's lifetime; the lock guards
--- a/src/protocol_bridge.cpp
+++ b/src/protocol_bridge.cpp
@@ -1,0 +1,207 @@
+#include "plugin.h"
+
+#include <charconv>
+
+using json = nlohmann::json;
+
+namespace {
+
+uint64_t asStreamId(const json& v) {
+    if (v.is_number_unsigned()) return v.get<uint64_t>();
+    if (v.is_number_integer()) return 0;
+    if (v.is_string()) {
+        const std::string& s = v.get_ref<const std::string&>();
+        uint64_t out = 0;
+        auto [p, ec] = std::from_chars(s.data(), s.data() + s.size(), out);
+        if (ec == std::errc{} && p == s.data() + s.size()) return out;
+    }
+    return 0;
+}
+
+bool parseBlob(const std::string& argsJson, const char* what, json& out, StdLogosResult& err) {
+    out = json::parse(argsJson, nullptr, false);
+    if (!out.is_object()) {
+        err = {false, {}, std::string(what) + ": invalid JSON object arg"};
+        return false;
+    }
+    return true;
+}
+
+constexpr uint64_t kDefaultReadMax = 1u << 20;
+
+uint64_t asReadMax(const json& a) {
+    auto it = a.find("maxSize");
+    if (it == a.end() || !it->is_number_unsigned()) return kDefaultReadMax;
+    uint64_t v = it->get<uint64_t>();
+    return v == 0 ? kDefaultReadMax : v;
+}
+
+}  // namespace
+
+StdLogosResult Libp2pModuleImpl::protocolRequest(const std::string& argsJson) {
+    json a;
+    StdLogosResult err;
+    if (!parseBlob(argsJson, "protocolRequest", a, err)) return err;
+
+    std::string peerId, proto, requestB64;
+    std::vector<std::string> multiaddrs;
+    int64_t timeoutMs = 0;
+    uint64_t maxSize = kDefaultReadMax;
+    bool expectResponse = true;
+    try {
+        peerId = a.at("peerId").get<std::string>();
+        proto = a.at("proto").get<std::string>();
+        requestB64 = a.at("requestB64").get<std::string>();
+        if (a.contains("multiaddrs"))
+            for (const auto& m : a["multiaddrs"]) multiaddrs.push_back(m.get<std::string>());
+        timeoutMs = a.value("timeoutMs", static_cast<int64_t>(0));
+        maxSize = asReadMax(a);
+        expectResponse = a.value("expectResponse", true);
+    } catch (...) {
+        return {false, {},
+                "protocolRequest: bad args (need {peerId,proto,multiaddrs?,requestB64,timeoutMs?,maxSize?,expectResponse?})"};
+    }
+
+    std::string requestBytes;
+    try {
+        requestBytes = base64Decode(requestB64);
+    } catch (const std::exception& e) {
+        return {false, {}, std::string("protocolRequest: bad requestB64: ") + e.what()};
+    }
+
+    if (!multiaddrs.empty()) {
+        auto c = connectPeer(peerId, multiaddrs, timeoutMs > 0 ? timeoutMs : kDefaultOpTimeoutMs);
+        if (!c.success) return {false, {}, "protocolRequest: connect failed: " + c.error};
+    }
+
+    auto d = dial(peerId, proto);
+    if (!d.success) return {false, {}, "protocolRequest: dial failed: " + d.error};
+    uint64_t streamId = 0;
+    try { streamId = d.value.get<uint64_t>(); } catch (...) {}
+    if (streamId == 0) return {false, {}, "protocolRequest: dial returned no stream"};
+
+    auto w = streamWriteLp(streamId, requestBytes);
+    if (!w.success) {
+        streamRelease(streamId);
+        return {false, {}, "protocolRequest: write failed: " + w.error};
+    }
+
+    if (!expectResponse) {
+        streamCloseWithEOF(streamId);
+        streamRelease(streamId);
+        return {true, json::object(), ""};
+    }
+
+    auto r = callSyncStreamWith(streamId, "Failed to read LP from stream",
+        [&](libp2p_stream_t* s, SyncPromise* p) {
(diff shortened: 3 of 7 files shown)


src/custom_handlers.cpp
```diff
@@ -34,9 +47,10 @@ void Libp2pModuleImpl::mountCompleteCallback(int ret, const char* msg, size_t le
 StdLogosResult Libp2pModuleImpl::mountProtocol(const std::string& proto) {
     if (!ctx) return {false, {}, "No libp2p context"};
     if (proto.empty()) return {false, {}, "Protocol string is empty"};
-    // Without emitEvent, protocolHandler would register a stream that no caller
-    // could ever read, close, or release — leaking the stream.
-    if (!emitEvent) return {false, {}, "emitEvent must be set before mounting a protocol"};
+    // Inbound streams are always reachable: protocolHandler queues each one for
+    // protocolAcceptStream (polling), and additionally emits protocolStream when
+    // emitEvent is set (in-process consumers). No emitEvent gate — a consumer
+    // MODULE drains the queue across the boundary, where push-events don't reach.
```
gmelodie:
```suggestion
```

src/plugin.h
```diff
@@ -149,6 +149,30 @@ class Libp2pModuleImpl {
...
+    //                          timeoutMs?, maxSize?} -> {responseB64}
+    //                         (connect + dial + writeLp + readLp + release).
+    //   streamReadLpJson:     {streamId, maxSize?, timeoutMs?} -> {dataB64}
+    //   streamWriteLpJson:    {streamId, dataB64} -> {}
+    //   streamCloseJson:      {streamId}
+    //   streamReleaseJson:    {streamId}
+    //   protocolAcceptStream: {proto, timeoutMs?} -> {streamId, proto}
+    //                         (polling drain of inbound streams for a mounted
+    //                         proto — protocolStream push-events don't cross the
+    //                         module boundary; mirrors gossipsubNextMessage).
```
gmelodie:
```suggestion
```

src/plugin.h
```diff
@@ -233,6 +257,16 @@ class Libp2pModuleImpl {
     std::condition_variable m_queueCond;
     std::unordered_map<std::string, std::queue<std::string>> m_topicQueues;
 
+    // Inbound custom-protocol streams accepted by a mounted handler, queued
+    // per-proto for polling delivery via protocolAcceptStream. protocolStream
+    // push-events don't cross the module boundary (the universal codegen skips
+    // the emitEvent sink), so a consumer MODULE drains accepted streams here —
+    // same shape as m_topicQueues above. Each entry is a JSON object
+    // {streamId, proto}.
```
gmelodie:
```suggestion
```

src/protocol_bridge.cpp
```diff
@@ -0,0 +1,192 @@
+#include "plugin.h"
+
+#include <cstdlib>
+
+using json = nlohmann::json;
+
+namespace {
+
+// Reads a streamId that may arrive as a JSON number or a decimal string (JSON
+// numbers past 2^53 lose precision, so a cautious caller may send a string).
```
gmelodie:
```suggestion
// Reads a streamId that may arrive as a JSON number or a decimal string
```

tests/custom_handlers.cpp
```diff
@@ -19,12 +20,14 @@ static std::pair<std::string, std::vector<std::string>> getPeerInfoPair(Libp2pMo
 
 static auto noopEmitEvent = [](const std::string&, const std::string&) {};
 
-LOGOS_TEST(custom_handlers_mount_requires_emit_event) {
+LOGOS_TEST(custom_handlers_mount_without_emit_event_ok) {
+    // Inbound streams are reachable via protocolAcceptStream (polling), so
+    // mounting no longer requires an in-process emitEvent sink — a consumer
+    // module that can't receive push-events across the boundary can still mount
+    // and serve.
```
gmelodie:
```suggestion
```

tests/custom_handlers.cpp
```diff
@@ -106,3 +109,65 @@ LOGOS_TEST(custom_handlers_protocol_stream_event) {
     LOGOS_ASSERT_TRUE(nodeA.stop().success);
     LOGOS_ASSERT_TRUE(nodeB.stop().success);
 }
+
+LOGOS_TEST(protocol_bridge_request_accept_roundtrip) {
+    // End-to-end over the generic bridge, using ONLY the JSON-blob methods a
+    // consumer module would call — no emitEvent, no typed dial/stream* calls.
+    // nodeB serves by polling protocolAcceptStream then streamReadLpJson /
+    // streamWriteLpJson; nodeA drives one protocolRequest and gets the echoed
+    // frame back. This is the cross-module usage shape exercised in-process.
```
gmelodie:
```suggestion
```

src/protocol_bridge.cpp
```diff
@@ -0,0 +1,192 @@
...
+uint64_t asStreamId(const json& v) {
+    if (v.is_number_unsigned()) return v.get<uint64_t>();
+    if (v.is_number_integer()) return static_cast<uint64_t>(v.get<int64_t>());
+    if (v.is_string()) return std::strtoull(v.get<std::string>().c_str(), nullptr, 10);
+    return 0;
+}
```
gmelodie, line 19: 1. no need for the integer/unsigned difference
2. `0` is a legal stream id, so we can't return that upon erroring

this is a better suggestion from claude 
```suggestion
#include <expected>
#include <charconv>

enum class StreamIdError { WrongType, NotANumber, Overflow };

[[nodiscard]] std::expected<uint64_t, StreamIdError>
asStreamId(const json& v) {
  if (v.is_number_integer())  // covers signed + unsigned
    return v.get<uint64_t>();

  if (!v.is_string())
    return std::unexpected(StreamIdError::WrongType);

  const auto& s = v.get_ref<const std::string&>();
  uint64_t out{};
  const auto* end = s.data() + s.size();
  auto [p, ec] = std::from_chars(s.data(), end, out);

  if (ec == std::errc::result_out_of_range)
    return std::unexpected(StreamIdError::Overflow);
  if (ec != std::errc{} || p != end)
    return std::unexpected(StreamIdError::NotANumber);
  return out;
}
```

src/protocol_bridge.cpp
```diff
@@ -0,0 +1,192 @@
...
+        requestBytes = base64Decode(requestB64);
+    } catch (const std::exception& e) {
+        return {false, {}, std::string("protocolRequest: bad requestB64: ") + e.what()};
+    }
+
+    // Dial needs the addr in the peerstore; connect first when addrs are given.
+    if (!multiaddrs.empty()) {
+        auto c = connectPeer(peerId, multiaddrs, timeoutMs > 0 ? timeoutMs : kDefaultOpTimeoutMs);
+        if (!c.success) return {false, {}, "protocolRequest: connect failed: " + c.error};
+    }
```
gmelodie: I wonder if we need to do that, won't dial do it for us?

src/protocol_bridge.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    auto d = dial(peerId, proto);
+    if (!d.success) return {false, {}, "protocolRequest: dial failed: " + d.error};
+    uint64_t streamId = 0;
+    try { streamId = d.value.get<uint64_t>(); } catch (...) {}
+    if (streamId == 0) return {false, {}, "protocolRequest: dial returned no stream"};
```
gmelodie, line 81: `0` can be a valid streamId i think

src/protocol_bridge.cpp
```diff
@@ -0,0 +1,192 @@
...
+    // The response frame may arrive only after a slow server round-trip (e.g. an
+    // on-chain write), so the read await tracks the caller's timeout, not the
+    // 10s default.
+    auto r = callSyncStreamWith(streamId, "Failed to read LP from stream",
+        [&](libp2p_stream_t* s, SyncPromise* p) {
+            return libp2p_stream_readLp(ctx, s, maxSize,
+                                        &Libp2pModuleImpl::promiseBufferCallback, p);
+        },
+        bufferToResult, awaitTimeoutFor(timeoutMs));
+    streamRelease(streamId);
```
gmelodie, line 101: What if your protocol does not expect a response?

src/protocol_bridge.cpp
```diff
@@ -0,0 +1,192 @@
...
+    uint64_t streamId = asStreamId(a["streamId"]);
+    uint64_t maxSize = a.value("maxSize", kDefaultReadMax);
+    int64_t timeoutMs = a.value("timeoutMs", static_cast<int64_t>(0));
+
+    auto r = callSyncStreamWith(streamId, "Failed to read LP from stream",
+        [&](libp2p_stream_t* s, SyncPromise* p) {
+            return libp2p_stream_readLp(ctx, s, maxSize,
+                                        &Libp2pModuleImpl::promiseBufferCallback, p);
+        },
+        bufferToResult, awaitTimeoutFor(timeoutMs));
```
gmelodie, line 130: `callSyncStreamWith` could be refactored to receive the timeout and run this closure inside so that calling places like this one would be something like `callSyncStreamWith(streamId, "Failed to read LP from stream", timeoutMs)`

src/custom_handlers.cpp
```diff
@@ -19,7 +19,20 @@ void Libp2pModuleImpl::protocolHandler(
...
     j["proto"] = protoStr;
-    self->emitEventSafe("protocolStream", j.dump());
+    const std::string payload = j.dump();
+
+    // Queue for polling delivery (protocolAcceptStream) — the emitEvent push
+    // sink doesn't cross the module boundary, so a consumer MODULE drains
+    // accepted streams from here, exactly as gossipsub does with topic queues.
+    {
+        std::lock_guard<std::mutex> lock(self->m_inboundStreamMutex);
+        self->m_inboundStreamQueues[protoStr].push(payload);
```
richard-ramos: m_inboundStreamQueues as it is right now can grow with stale streams,

normally when you receive the streamId via that event, you'd use the stream handling API, and finally call `streamRelease` to get rid of the stream,  however, that does not remove the stream from m_inboundStreamQueues... thus doing a poll later might end up with a stale stream being returned

src/protocol_bridge.cpp
```diff
@@ -0,0 +1,192 @@
...
+    int64_t timeoutMs = 0;
+    uint64_t maxSize = kDefaultReadMax;
+    try {
+        peerId = a.at("peerId").get<std::string>();
+        proto = a.at("proto").get<std::string>();
+        requestB64 = a.value("requestB64", std::string());
+        if (a.contains("multiaddrs"))
+            for (const auto& m : a["multiaddrs"]) multiaddrs.push_back(m.get<std::string>());
+        timeoutMs = a.value("timeoutMs", static_cast<int64_t>(0));
+        maxSize = a.value("maxSize", kDefaultReadMax);
```
richard-ramos: Should these inputs be validated? to make sure they contain a valid int64 and uint64?

src/protocol_bridge.cpp
adklempner: `dial(peerId, proto)` takes no multiaddr, it opens a stream on an already-known peer

`connectPeer(peerId, multiaddrs)` is what establishes the connection from the addrs, and it's guarded by `if (!multiaddrs.empty())`, so an already-connected caller skips it

src/protocol_bridge.cpp
adklempner: not sure that `0` is valid

ids start at 1 (`m_nextStreamId{1}`), and dial itself returns `{true, 0}` when there's no stream

src/protocol_bridge.cpp
adklempner: added optional `expectResponse (default true);`

when false it half-closes (EOF) and returns after the write, so fire-and-forget protocols work. New test `protocol_bridge_request_no_response` covers it.

src/custom_handlers.cpp
adklempner: ` m_inboundStreamQueues[proto]` is now bounded (1024), let me know if that number makes sense

src/protocol_bridge.cpp
adklempner: the `protocolRequest` arg block is inside a try/catch, so on wrong types it should thrown and return a "bad args" error

`streamReadLpJson` didn't have the block so I added it

src/protocol_bridge.cpp
adklempner: Took the `from_chars` idea but returning 0 rather than std::expected

every call site funnels the result into `callSyncStreamWith/streamRelease`, which already collapse to a uniform "Invalid stream", so 0 (the module's reserved invalid id) lands on the same path with no call-site churn

also folds in Copilot's point re negatives. my claude claims your claude's suggestion would still pass -1 through as 2^64-1

src/protocol_bridge.cpp
adklempner: it already takes a timeout as the very last arg

src/custom_handlers.cpp
richard-ramos: Hm, but bounding the queue while it does prevent unbounded growth, it does not really fix the issue i described here.

The problem is that If an user handles streams with  `protocolStream` event and then calls `streamRelease`, the stream is removed from `m_streams`, but not from `m_inboundStreamQueues`.

I think we need to remove the stream from `m_inboundStreamQueues` when it is released

src/custom_handlers.cpp
```diff
@@ -2,6 +2,10 @@
 
 using json = nlohmann::json;
 
+// Cap on un-drained inbound streams per protocol: backpressure so a peer that
+// opens streams nobody polls can't grow m_streams/the queue without bound.
```
gmelodie:
```suggestion
```

src/custom_handlers.cpp
```diff
@@ -19,7 +23,31 @@ void Libp2pModuleImpl::protocolHandler(
...
+            queued = true;
+        }
+    }
+    if (!queued) {
+        self->removeStream(streamId);
+        libp2p_stream_release(self->ctx, stream,
+                              +[](int, const char*, size_t, void*) {}, nullptr);
+        return;
+    }
+    self->m_inboundStreamCond.notify_all();
```
gmelodie: Please extract this into a separate function to make this simpler and avoid the long comment.
 
```cpp
bool Host::enqueueInboundStream(const std::string& proto,
                                const InboundStreamPayload& payload,
                                libp2p_stream_t* stream,
                                StreamId streamId) {
    bool queued = false;
    {
        std::lock_guard<std::mutex> lock(m_inboundStreamMutex);
        auto& q = m_inboundStreamQueues[proto];
        if (q.size() < kMaxInboundStreamsPerProto) {
            q.push(payload);
            queued = true;
        }
    }

    if (!queued) {
        removeStream(streamId);
        libp2p_stream_release(ctx, stream,
                              +[](int, const char*, size_t, void*) {}, nullptr);
        return false;
    }

    m_inboundStreamCond.notify_all();
    return true;
}
```

Then call it like

```cpp
if (!self->enqueueInboundStream(protoStr, payload, stream, streamId)) {
    return;
}
```

src/protocol_bridge.cpp
richard-ramos: would it make sense to validate the uint64 value so it's >= 0?

src/protocol_bridge.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (v.is_number_unsigned()) return v.get<uint64_t>();
+    if (v.is_number_integer()) return 0;  // negative
+    if (v.is_string()) {
+        const std::string& s = v.get_ref<const std::string&>();
+        uint64_t out = 0;
+        auto [p, ec] = std::from_chars(s.data(), s.data() + s.size(), out);
+        if (ec == std::errc{} && p == s.data() + s.size()) return out;
+    }
+    return 0;
+}
```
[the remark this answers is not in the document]
richard-ramos: Dunno if true, but sounds like a valid concern

src/protocol_bridge.cpp
```diff
@@ -0,0 +1,211 @@
...
+    try {
+        dataBytes = base64Decode(a.value("dataB64", std::string()));
+    } catch (const std::exception& e) {
+        return {false, {}, std::string("streamWriteLpJson: bad dataB64: ") + e.what()};
+    }
+
+    auto w = streamWriteLp(streamId, dataBytes);
+    if (!w.success) return w;
+    return {true, json::object(), ""};
+}
```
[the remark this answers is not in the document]
richard-ramos: :thinking:

src/protocol_bridge.cpp
gmelodie: Ah! I see what's going on. In nim-libp2p there are multiple definitions of `dial`, one being `dial(peerId, addrs, protos, forceDial)`. This is important because if `connect` succeeds but `dial` fails we strand the connection, while upstream's `dial` (with `addrs`) handles that teardown.

src/protocol_bridge.cpp
gmelodie: Aha! The confusion on my end was between this `stream` and nim-libp2p's muxers' `stream`s (which in fact can be zero). Wdyt of adding/using this:

```cpp
static constexpr uint64_t kInvalidStreamId = 0;
```

src/plugin.h
```diff
@@ -149,6 +149,22 @@ class Libp2pModuleImpl {
...
 
+    // Generic custom-protocol bridge: opaque base64 payloads let a consumer
+    // module drive any length-prefixed protocol over the module boundary (the
+    // typed dial/stream* signatures don't marshal through the universal codegen).
+    //   protocolRequest:      {peerId, multiaddrs?, proto, requestB64, timeoutMs?,
+    //                          maxSize?, expectResponse?} -> {responseB64}
+    //   streamReadLpJson:     {streamId, maxSize?, timeoutMs?} -> {dataB64}
+    //   streamWriteLpJson:    {streamId, dataB64} -> {}
+    //   streamCloseJson / streamReleaseJson: {streamId}
+    //   protocolAcceptStream: {proto, timeoutMs?} -> {streamId, proto}  (polling)
```
gmelodie:
```suggestion
```

Image

src/protocol_bridge.cpp
gmelodie: https://github.com/logos-co/logos-libp2p-module/issues/93

gmelodie (commented):
Apologies for re-reviewing parts of the code, sometimes a second pass reveals more stuff. Mostly minor though :) Thank you for doing this

gmelodie (approved):

src/protocol_bridge.cpp
adklempner: added a check for `maxSize`, `timeoutMs` already gets clamped when <= 0 to a default

src/protocol_bridge.cpp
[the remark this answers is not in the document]
adklempner: updated so missing field returns an error, explicit `""` returns a valid empty frame

richard-ramos (approved):

PR #84, opened 2026-07-23
docs(tutorial): add tutorials 5, 6, 7, 8

vladopajic:
part 3 of #78 

just adding tutorials 5, 6, 7 (7 is split into two 7 and 8)

### DIFF
--- a/tutorial/docs/tutorial_5_kademlia_basics.md
+++ b/tutorial/docs/tutorial_5_kademlia_basics.md
@@ -2,14 +2,12 @@
 
 Kademlia is a Distributed Hash Table (DHT) that lets peers store and
 retrieve values without a central server. The `logos-libp2p-module`
-exposes this through the `kadPutValue()`, `kadGetValue()`, and
-`kadGetRandomRecords()` functions.
+exposes this through the `kadPutValue()` and `kadGetValue()` functions.
 
 In this tutorial we'll:
   - Start two nodes that bootstrap to each other
   - Store a value from one node
   - Retrieve it from the other node
-  - List random records to verify DHT participation
 
 ## How Kademlia works in libp2p
 
@@ -22,7 +20,8 @@ certain "keys" (hashes). When you call `kadPutValue(key, value)`:
 When you call `kadGetValue(key, quorum)`:
   1. The key is hashed and closest peers are looked up
   2. The value is requested from them
-  3. The quorum parameter controls how many responses are needed
+  3. The quorum parameter controls how many peers must return the same
+     value for the key before the lookup succeeds
 
 ## Bootstrap nodes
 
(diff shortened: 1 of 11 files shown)


richard-ramos (approved):
A-mazing!

tutorial/docs/tutorial_5_kademlia_basics.md
```diff
@@ -0,0 +1,189 @@
...
+#include <cstdio>
+#include <string>
+#include <vector>
+#include "plugin.h"
+
+int main()
+{
+    printf("=== Tutorial 5: Kademlia DHT Basics ===\n\n");
+
+```
```
gmelodie, line 40: Still don't understand why this snippet is here, seems like it has nothing to do with what's being said

tutorial/docs/tutorial_5_kademlia_basics.md
```diff
@@ -0,0 +1,189 @@
+# Tutorial 5: Kademlia DHT Basics
```
gmelodie, line 1: For this tutorial, it'd be cool to have more nodes and have A store a key in B and have C retrieve the key (maybe from D).

tutorial/docs/tutorial_6_kademlia_providers.md
```diff
@@ -0,0 +1,221 @@
...
+#include <cstdio>
+#include <string>
+#include <vector>
+#include "plugin.h"
+
+int main()
+{
+    printf("=== Tutorial 6: Kademlia Provider Records ===\n\n");
+
+```
```
gmelodie, line 36: same comment as before

tutorial/docs/tutorial_7_gossipsub_polling.md
```diff
@@ -0,0 +1,294 @@
...
+#include <thread>
+#include <string>
+#include <vector>
+#include "plugin.h"
+
+int main()
+{
+    printf("=== Tutorial 7: GossipSub – Pub/Sub Messaging ===\n\n");
+
+```
```
gmelodie, line 37: same comment as before

tutorial/docs/tutorial_7_gossipsub.md
```diff
@@ -0,0 +1,294 @@
...
+    printf("Connected\n");
+
+```
+
+## Step 3: Prepare an event callback for Node A
+
+This callback is only needed for Step 8, where Node A receives a
+message through the event-driven API. The polling flow in Steps 4-7
+uses `gossipsubNextMessage()` instead and does not depend on this
+callback.
```
gmelodie: I think it would make sense to have two source code files for this single tutorial, one using `gossipsubNextMessage`, and the other using events. This way we'd avoid the huge gap between steps 3 and 8 in this doc. Would make the examples easier to follow

gmelodie (commented):
Looking good except for the two gossipsub source files suggestion

tutorial/docs/tutorial_5_kademlia_basics.md
vladopajic: what snippet?

tutorial/docs/tutorial_5_kademlia_basics.md
vladopajic: if i am not mistaking, that will not increase / improve knowledge passing to reader, and will just complicate example. 

point of that tutorial is to cover minimal interface how to use kademlia. many things are not really said nor demonstrated, but this is not supposed to. for example if value will passed via connected node, when quorum is not reach, etc...

tutorial/docs/tutorial_7_gossipsub.md
vladopajic: yes makes sense. i was thinking about it but then waited to see if reviews would prefer the same

PR #85, opened 2026-07-23
ci(tutorials): add smoke test for code

vladopajic:
previously tutorials CI was just making sure that tutorials code compiles. 

with this change CI will execute tutorial code preforming smoke tests on them. all tutorials should finish with success. unsuccessful outcome can happen when, for example:

```cpp
    // faild to dial
    if (!dialRes.success) {
        fprintf(stderr, "Dial failed: %s\n", dialRes.error.c_str());
        return 1;
    }
```

or 

```cpp
        // value stored and recived form kad is not the same,
        if (received == value) {
            printf("Value matches!\n");
        } else {
            fprintf(stderr, "Value mismatch (expected: \"%s\")\n",
                   value.c_str());
            return 1;
        }
```        

etc ...

----

improving tutorial code, handling errors first. this should be general principle of any code

### DIFF
--- a/.github/workflows/tutorial.yml
+++ b/.github/workflows/tutorial.yml
@@ -56,6 +56,36 @@ jobs:
         run: |
           nix develop --command ./tutorial/build_tutorials.sh
 
+      # Run tutorials after compilation as a simple check that the tutorial code is also correct.
+      - name: Run tutorial examples inside Nix dev shell
+        run: |
+          nix develop --command bash -c '
+            set -euo pipefail
+            shopt -s nullglob
+
+            tutorial_binaries=(./build/tutorial/tutorial_*)
+            if [[ ${#tutorial_binaries[@]} -eq 0 ]]; then
+              echo "No tutorial binaries found in ./build/tutorial/"
+              exit 1
+            fi
+
+            ran_any=0
+            for tutorial_binary in "${tutorial_binaries[@]}"; do
+              if [[ ! -f "${tutorial_binary}" || ! -x "${tutorial_binary}" ]]; then
+                continue
+              fi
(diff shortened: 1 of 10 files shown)


.github/workflows/tutorial.yml
```diff
@@ -56,6 +56,36 @@ jobs:
...
+              echo "Running ${tutorial_binary}"
+              "${tutorial_binary}"
+              ran_any=1
+            done
+
+            if [[ "${ran_any}" -eq 0 ]]; then
+              echo "No executable tutorial binaries found in ./build/tutorial/"
+              exit 1
+            fi
+          '
```
gmelodie, line 87: nit: i wonder if this should be a script inside a file

tutorial/build_tutorials.sh
```diff
@@ -66,8 +66,40 @@ if [[ "${#tutorial_targets[@]}" -eq 0 ]]; then
...
+        echo "Patching ${plugin_dylib} libp2p.dylib dependency"
+        install_name_tool -change \
+            "${libp2p_install_name}" \
+            "@loader_path/libp2p.dylib" \
+            "${plugin_dylib}"
+    fi
+fi
 
 echo ""
 echo "✅ All tutorials compiled successfully."
```
gmelodie, line 105:
```suggestion
echo "All tutorials compiled successfully."
```

gmelodie (approved):
cool! only nits

richard-ramos (approved):

tutorial/build_tutorials.sh
vladopajic: this is intentional

PR #90, opened 2026-07-24
docs(tutorial): add introduction tutorial

vladopajic:
- add tutorial 0 - an introduction tutorial
- use `StdLogosResult` instead of `auto` - this is much more readable for novice users
- avoid checking json value types, before casting 
    - makes tutorials more focused
    - if there is mismatch, compile or tutorial smoke tests should fail


gmelodie (approved):

tutorial/tutorial_0_introduction.cpp
```diff
@@ -0,0 +1,149 @@
...
+/// This keeps the code predictable: errors are handled immediately, and the
+/// rest of the step can assume the operation succeeded.
+///
+/// ## Tutorial Executables Fail Fast
+///
+/// These programs are examples, not long-running services. When a required
+/// operation fails, they print a useful error message to `stderr` and return
+/// `1`. Successful tutorials print progress to `stdout` and return `0`.
+///
+/// The `stdout` stream can be noisy because logs from the wrapped C binding
```
richard-ramos, line 43:
```suggestion
/// The `stdout` stream can be noisy because logs from the wrapped C binding are also included
```

tutorial/tutorial_0_introduction.cpp
```diff
@@ -0,0 +1,149 @@
...
+                version.error.c_str());
+        return 1;
+    }
+    printf("Module version: %s\n",
+           version.value.get<std::string>().c_str());
+
+/// ## Step 3: Stop cleanly
+///
+/// For production code it should always be better to attempt to gracefully
+/// close all resources first.
```
richard-ramos:
```suggestion
/// Production code should always attempt to close resources gracefully first
```

tutorial/tutorial_1_node_lifecycle.cpp
richard-ramos: Since there's a Tutorial 0 now, should this still say "Welcome to the first tutorial?"

tutorial/docs/tutorial_0_introduction.md
```diff
@@ -0,0 +1,164 @@
...
+    if (!version.success) {
+        fprintf(stderr, "Failed to get module version: %s\n",
+                version.error.c_str());
+        return 1;
+    }
+    printf("Module version: %s\n",
+           version.value.get<std::string>().c_str());
+
+```
+
```
richard-ramos, line 115: Would be interesting to have also a small tutorial that teachs how to interact with non scalar values, like  peerInfo() which can contain multiple addresses

tutorial/docs/tutorial_1_node_lifecycle.md
richard-ramos, line 56: To follow the suggestion from tutorial 0 to print the error on !success, we should do that here too

richard-ramos (approved):

PR #95, opened 2026-07-29
feat: logs filter

vladopajic:
- adds log filter 
- all tutorials disable filter to show clean output

closes: #79

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -25,7 +25,7 @@
         value = f system;
       }) systems);
 
-      libp2pRuntimeFiltering = {
+      libp2pInputs = {
         packages = forEachSystem (system: {
           cbind = inputs.libp2p.packages.${system}.cbind.overrideAttrs (old: {
             buildPhase = builtins.replaceStrings
@@ -38,7 +38,7 @@
 
       externalLibInputs = {
         libp2p = {
-          input = libp2pRuntimeFiltering;
+          input = libp2pInputs;
           packages.default = "cbind";
         };
       };
--- a/flake.nix
+++ b/flake.nix
@@ -3,7 +3,7 @@
 
   inputs = {
     logos-module-builder.url = "github:logos-co/logos-module-builder";
-    libp2p.url = "github:vacp2p/nim-libp2p/fix/cbind/nim-ffi-bump";
+    libp2p.url = "github:vacp2p/nim-libp2p/master";
 
     openmetrics-module = {
       url = "github:logos-co/openmetrics-module";
--- a/flake.nix
+++ b/flake.nix
@@ -17,9 +17,28 @@
 
   outputs = inputs@{ logos-module-builder, ... }:
     let
+      nixpkgs = logos-module-builder.inputs.nixpkgs;
+      systems = [ "aarch64-darwin" "x86_64-darwin" "aarch64-linux" "x86_64-linux" ];
+
+      forEachSystem = f: builtins.listToAttrs (map (system: {
+        name = system;
+        value = f system;
+      }) systems);
+
+      libp2pRuntimeFiltering = {
+        packages = forEachSystem (system: {
+          cbind = inputs.libp2p.packages.${system}.cbind.overrideAttrs (old: {
+            buildPhase = builtins.replaceStrings
+              [ "--threads:on --opt:size --noMain --mm:refc --d:metrics" ]
+              [ "--threads:on --opt:size --noMain --mm:refc --d:metrics -d:chronicles_runtime_filtering=on" ]
+              old.buildPhase;
+          });
+        });
+      };
+
       externalLibInputs = {
         libp2p = {
-          input = inputs.libp2p;
+          input = libp2pRuntimeFiltering;
           packages.default = "cbind";
         };
       };
@@ -33,14 +52,6 @@
           dir = ./tests;
         };
       };
-
-      nixpkgs = logos-module-builder.inputs.nixpkgs;
-      systems = [ "aarch64-darwin" "x86_64-darwin" "aarch64-linux" "x86_64-linux" ];
-
-      forEachSystem = f: builtins.listToAttrs (map (system: {
-        name = system;
-        value = f system;
-      }) systems);
 
       # Pre-resolved store paths for the two .lgx bundles the e2e installs.
       e2eEnv = system: {
--- a/src/plugin.cpp
+++ b/src/plugin.cpp
@@ -214,13 +214,9 @@ StdLogosResult Libp2pModuleImpl::createNode(const std::string& config) {
 }
 
 StdLogosResult Libp2pModuleImpl::setLogLevel(LogLevel level) {
-    if (!ctx) {
-        auto created = createContext();
-        if (!created.success) return created;
-    }
     return callSync("Failed to set log level", [&](SyncPromise* p) {
-        return libp2p_ctx_set_log_level(ctx, static_cast<int64_t>(level),
-                                        &Libp2pModuleImpl::cbBool, p);
+        return libp2p_static_set_log_level(static_cast<int64_t>(level),
+                                           &Libp2pModuleImpl::cbBool, p);
     });
 }
 
--- a/src/plugin.cpp
+++ b/src/plugin.cpp
@@ -178,6 +178,7 @@ StdLogosResult Libp2pModuleImpl::createContext() {
     // identity; a supplied key gives a stable peer id across restarts.
     m_libp2pConfig.privKey =
         NimFfiBytes{m_privKey.empty() ? nullptr : m_privKey.data(), m_privKey.size()};
+    m_libp2pConfig.logLevel = g_requestedLogLevel.load();
 
     auto r = spawnContext(m_libp2pConfig);
     if (!r.ok) {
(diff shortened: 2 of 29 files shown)


src/plugin.h
```diff
@@ -53,6 +53,17 @@ struct SyncResult {
     LibP2PCtx* newCtx = nullptr;
 };
 
+enum class LogLevel : int64_t {
+    None = 0,
+    Trace = 1,
+    Debug = 2,
+    Info = 3,
+    Notice = 4,
```
gmelodie: wtf is `notice`

src/plugin.cpp
```diff
@@ -213,6 +213,24 @@ StdLogosResult Libp2pModuleImpl::createNode(const std::string& config) {
...
+    return callSync("Failed to set log level", [&](SyncPromise* p) {
+        return libp2p_ctx_set_log_level(ctx, static_cast<int64_t>(level),
+                                        &Libp2pModuleImpl::cbBool, p);
+    });
+#else
+    // Older generated cbind headers do not expose runtime log filtering.
+    // Keep this a no-op so callers can use one source path across cbind revs.
+    (void)level;
+    return {true, {}, ""};
+#endif
```
gmelodie: Since we don't have a stable release yet, I don't think we need to worry about breaking the public api
```suggestion
```

tutorial/docs/tutorial_0_introduction.md
```diff
@@ -40,8 +40,22 @@ These programs are examples, not long-running services. When a required
...
+level such as `LogLevel::Info`, `LogLevel::Debug`, or `LogLevel::Trace` when
+you need more detail while debugging:
+
+```cpp
+StdLogosResult logRes = node.setLogLevel(LogLevel::None);
+if (!logRes.success) {
+    fprintf(stderr, "Failed to disable libp2p logs: %s\n",
+            logRes.error.c_str());
+    return 1;
+}
```
gmelodie: If we get rid of the `#ifdef..#endif` stuff from above:

```suggestion
node.setLogLevel(LogLevel::None);
```

flake.nix
```diff
@@ -3,7 +3,7 @@
 
   inputs = {
     logos-module-builder.url = "github:logos-co/logos-module-builder";
-    libp2p.url = "github:vacp2p/nim-libp2p/fix/cbind/nim-ffi-bump";
+    libp2p.url = "github:vacp2p/nim-libp2p/master";
```
gmelodie, line 6: don't fix this to master, as changes can spill here unexpectedly

```suggestion
    libp2p.url = "github:vacp2p/nim-libp2p/fix/cbind/nim-ffi-bump";
```
or the commit that was merged in master
```suggestion
    libp2p.url = "github:vacp2p/nim-libp2p/a760255635529adbc2e8f198b8b53ba13396c4ec";
```

gmelodie (commented):
reviewing as i was almost done before it got converted back

flake.nix
vladopajic: changes can't spill here unexpectedly as it is fixed to master, with rev `a70c615c287a94e1134f45c05c18c338c7a3bc8f`.

vladopajic: see flake.lock file

src/plugin.h
vladopajic: https://github.com/status-im/nim-chronicles/blob/b8ca1cc63aa0ce037d07a15207efa6b71b86de5e/chronicles.nim#L390

vladopajic: it is 1:1 mapping to chronicles log lvl

tutorial/docs/tutorial_0_introduction.md
vladopajic: hmm i don't see what `#ifdef..#endif` has to do with this code? 

result from `setLogLevel` has to be checked because `cbind` returns `result`.

flake.nix
gmelodie: Yes! but when doing `nix flake update` to update some other dep this will also update nim-libp2p to latest master, which might not be what we want

tutorial/docs/tutorial_0_introduction.md
gmelodie: I guess the overall issue here is that set log level is creating `ctx`? Why is that. I would expect `setLogLevel` to never fail (i.e. do not create a context, require one and that's it).

This would keep us from doing 
```cpp
  if (!logRes.success) {
        fprintf(stderr, "Failed to disable libp2p logs: %s\n",
                logRes.error.c_str());
        return 1;
    }
```
every time

vladopajic: https://github.com/vacp2p/nim-libp2p/blob/a70c615c287a94e1134f45c05c18c338c7a3bc8f/cbind/libp2p.nim#L417

^ this is `setLogLevel` from cbind. that `setLogLevel` returns results, so this result has to be checked. this is what we are doing here when checking result.

flake.nix
vladopajic: why would it be okay for other deps and not for `nim-libp2p` dep?

src/plugin.cpp
```diff
@@ -213,6 +213,17 @@ StdLogosResult Libp2pModuleImpl::createNode(const std::string& config) {
...
 }
 
+StdLogosResult Libp2pModuleImpl::setLogLevel(LogLevel level) {
+    if (!ctx) {
+        auto created = createContext();
+        if (!created.success) return created;
+    }
+    return callSync("Failed to set log level", [&](SyncPromise* p) {
+        return libp2p_ctx_set_log_level(ctx, static_cast<int64_t>(level),
+                                        &Libp2pModuleImpl::cbBool, p);
```
richard-ramos: @gmelodie with nim-ffi, is it always required to have a context? I am wondering if we could just update nim-libp2p `libp2pSetLogLevel` function so it looks like:

```nim
proc libp2pSetLogLevel*(level: int): Future[Result[bool, string]] {.ffi.} =
  ## Changes the process-wide Chronicles runtime log level.
  if level < ord(low(chronicles.LogLevel)) or level > ord(high(chronicles.LogLevel)):
    return err("invalid log level: " & $level)

  when chronicles.runtimeFilteringEnabled:
    chronicles.setLogLevel(chronicles.LogLevel(level))
    ok(true)
  else:
    err("Chronicles runtime filtering is disabled")
```

and that way you avoid allocating memory for context and also breaking createNode.

gmelodie: Yes it should be possible to just do that

flake.nix
gmelodie: Let's say we need to update logos core, but nim-libp2p has some breaking changes that have not been ported to cbind yet.

src/plugin.cpp
vladopajic: yeah, this will make things better. i'll push this to nim-libp2p

flake.nix
vladopajic: >  nim-libp2p has some breaking changes that have not been ported to cbind yet

could never happen. every push to nim-libp2p master has to pass CI. if nim-libp2p code brakes cbind CI that PR will not get through. 

ether way, even when we assume that, that case can happen, those are exceptional cases. meaning that by default we want default behavior, which is `nix flake update` to updated everything. and exceptional cases are dealt with an exception. in this hypothetical case, one would still do `nix flake update` and manually rivert changes related to nim-libp2p.

tutorial/docs/tutorial_0_introduction.md
```diff
@@ -40,8 +40,28 @@ These programs are examples, not long-running services. When a required
...
+level to `LogLevel::None` by default so `stdout` only shows the tutorial's
+own progress messages during normal runs. Choose another level such as
+`LogLevel::Info`, `LogLevel::Debug`, or `LogLevel::Trace` when you need more
+detail while debugging:
+
+In your own application, `LogLevel::Error` or `LogLevel::Fatal` is often a
+useful default. It keeps normal output quiet while still surfacing conditions
+that may indicate libp2p is misbehaving or that your integration code needs an
+adjustment. Some error logs describe remote-peer behavior, retries, or
+recoverable internal state, so they may not require any action from your side.
```
vladopajic: added this as a practical suggestion because users should not generally disable logs completely, especially while building or integrating an application.

still not fully sure this should be the final recommendation, because libp2p logs are not currently structured in a way that is consistently helpful to end users. there are cases where an error log may be written to stdout, but the condition is not necessarily actionable for the application. in some cases, the log level appears to be error mainly because the log statement is inside an `except SomeError` block.

second reason for adding this note is to raise the topic of improving (more accurately creating) a systematic approach to choosing log levels. runtime logging should be treated as a library feature with user-facing behavior. and now when we are not only a developers, but users as well, it may be easier for us to understand how to approach this.

this may also expand the scope of the commitment `ift-ts:p2p:ift:2026q3-nimlibp2p-log-noise-reduction:logging-changes`.

.... but it might be completely unnecessary in age of ai. but even with this assumption systematic approach would be useful to ai.

tutorial/docs/tutorial_0_introduction.md
vladopajic: pr [#2899](https://github.com/vacp2p/nim-libp2p/pull/2899) makes `setLogLevel` a static function. this change requires that call sites continue to check the result, despite avoiding the context.

vladopajic: there is no validation when setting log lvl anymore, therefore no result (response).
validation is done on cbind side before libp2p node is created.
in logos module it will be impossible to set invalid value, as user will have to use `LogLevel` enum.

src/plugin.cpp
vladopajic: LogLevel is now passed via config, avoiding need for context.

gmelodie (approved):

PR #96, opened 2026-07-31
chore: expose `libp2p_ctx_new_private_key`

vladopajic:


gmelodie (approved):

src/plugin.cpp
```diff
@@ -255,6 +255,22 @@ StdLogosResult Libp2pModuleImpl::status() {
     return {false, {}, m_initError.empty() ? "libp2p not initialized" : m_initError};
 }
 
+StdLogosResult Libp2pModuleImpl::newPrivateKey(KeyScheme scheme) {
+    if (!ctx) {
+        auto created = createContext();
+        if (!created.success) return created;
+    }
```
richard-ramos, line 262: It shouldnt be necessar to have a context just to create a private key.

vladopajic: it uses rng from context currently

richard-ramos: I see. This is bad because of how createNode works: https://github.com/logos-co/logos-libp2p-module/blob/5e67df57fbfcbdfba2fe799e7b6fec15e0674982/src/plugin.cpp#L207-L211

If you were to call libp2p_ctx_new_private_key BEFORE createNode (which should be probably the use case, as you'd want a private key to later use in the node config), createNode will fail with a `"createNode: node already created"`.

So, if a context is being created here, you gotta make sure you destroy it afterwards with destroyContext. Altho ideally a context should not be needed. I'll create an issue in nim-libp2p for tracking that part.

PR #97, opened 2026-07-31
fix(lidl): give createXpr services and newPrivateKey a LIDL type

dlipicar:
We're making interface definition for modules a bit stricter (things that we currently kind of automagically convert to something sendable across the module interface is gonna become a build error soon) to fix some cases where we had hard to find data losses on edge cases.

- Enums are not (currently) supported -> Changing `KeyScheme` with a string representation.
- std::pair is not a supported type for a module interface -> Replaced vector of pairs to map keyed by the first element of the pair (you tell me if multiple entries with the same ID are expected or the order matters, then map is not the proper data structure)
- std::vector<uint8_t> is the proper data type to losslessly transmit byte arrays without having to manually base64-stringify them first. But you're free to keep the base64 string if you wish.

### DIFF
--- a/src/callbacks.cpp
+++ b/src/callbacks.cpp
@@ -3,26 +3,29 @@
 using json = nlohmann::json;
 
 namespace {
-// {peerId, seqNo, addrs, services:{id: data}}. Keyed by service id, mirroring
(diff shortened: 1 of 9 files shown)


gmelodie (approved):
ty!

PR #99, opened 2026-08-06
fix(service-disco): advertise a caller-supplied XPR

gmelodie:
Closes #98.

`startAdvertising` now takes the advertisement to publish, so a consuming module that runs on its own switch advertises its own peer ID and addresses. Before, the module encoded `serviceInfo` into an XPR built from its own connectable address, and discoverers reached the wrong node.

`discoStartAdvertising(serviceId, serviceData, advertisement)`: `advertisement` is a base64 signed XPR, the same encoding `createXpr` returns and `decodeXpr` accepts. An empty string keeps the current behaviour, and the node advertises its own record. The module decodes the base64 and passes the bytes through; nim-libp2p's `validateAdvert` rejects a record that fails to decode, exceeds `MaxXPRSize`, or does not list the service, and the reason reaches the caller.

The `libp2p` input stays on `master` and moves to `00cab64`, the merge of vacp2p/nim-libp2p#2917, which adds the `advertisement: seq[byte]` field to `StartAdvertisingRequest` and makes `startAdvertising` return `Result[void, string]`.

That bump also picks up a week of nim-libp2p master, which renamed the generated entry points for `{.ffiStatic.}` procs. Two unrelated call sites follow the rename: `libp2p_ctx_collect_metrics` becomes `libp2p_static_collect_metrics`, and `libp2p_ctx_new_private_key` becomes `libp2p_static_new_private_key`. Both keep `callSyncWith`, so the existing context requirement does not change.

Tests:
- `disco_advertise_foreign_xpr` builds an XPR on a node that mounts no service discovery, advertises it from a second node, and asserts that the discoverer resolves the first node's peer ID and address.
- `disco_advertise_rejects_bad_xpr` covers bad base64, an XPR for a different service, and a tampered signature.
- The bootstrap-plus-two-peers fixture the disco tests repeated three times is now `startDiscoPeer`.

Verified locally: `nix build .#lgx` succeeds on the merged `master` pin. The integration suite ran earlier on the pre-merge branch pin, whose cbind API is identical, and reported 84 passed, 3 failed. All disco and XPR tests pass, including the two new ones.

Two pre-existing issues found while verifying, both out of scope here:
- CI never builds the integration suite. `tests/CMakeLists.txt` looks for `libp2p.so` in `../lib`, the test derivation does not vendor it, and every job logs `[Libp2pTests] libp2p not found in ../lib — skipping integration tests`. The suite has to be built by hand to run at all.
- The 3 failures are the consequence: `custom_handlers_mount_without_emit_event_ok`, `protocol_bridge_request_accept_roundtrip` and `protocol_bridge_request_no_response` all call `mountProtocol` on a node with no `emitEvent`, which `src/custom_handlers.cpp:9` rejects by design. The tests contradict the implementation and nothing has been running them.
- `createXpr` is missing from the generated LIDL contract because its declaration in `plugin.h` spans three lines and the header parser drops those.


vladopajic (approved):

flake.nix
```diff
@@ -3,7 +3,7 @@
 
   inputs = {
     logos-module-builder.url = "github:logos-co/logos-module-builder";
-    libp2p.url = "github:vacp2p/nim-libp2p/master";
+    libp2p.url = "github:vacp2p/nim-libp2p/fix/cbind/advertise-full-xpr";
```
vladopajic: this should not be committed; it is okay for local and testing but main/master branch of of `logos-libp2p-module` should never depend on experimental branches. 

this essentially means that this pr needs to wait for `fix/cbind/advertise-full-xpr` to hit master before merging.

because it also means that experimental branch has satisfied review and ci check. currently there is no guarantees when depending on experimental branches.

richard-ramos (approved):

PR #101, opened 2026-08-07
fix(gossipsub): bound the per-topic message queue and expose the ingress limits

gmelodie:
## Summary

Every GossipSub message on a subscribed topic went into a `std::queue` in `Libp2pModuleImpl::onPubsubMessage`. Only `gossipsubNextMessage()` drains that queue. An application that reads the `gossipsubMessage` event kept every message it ever received, for the lifetime of the module object.

That was the last unbounded hop in the chain. nim-libp2p bounds its per-peer outbound queues, nim-ffi bounds its event ring at 1024, and go-libp2p-pubsub bounds the equivalent subscription channel at 32 and drops the newest. This module drained a bounded ring into an unbounded queue.

The backlog is now a `TopicQueues` type (`src/topic_queues.h`, `src/topic_queues.cpp`). A message count and a total byte count bound it together. Both bounds are necessary, because 1024 messages at the 1 MiB GossipSub limit is still 1 GiB per topic. On overflow the newest message is dropped, so a slow consumer gets a coherent prefix of the stream and then a gap.

The queue bounds hold what a peer already delivered. This PR also exposes the nim-libp2p GossipSub ingress limits, which bound what a peer can deliver: `gossipsubMaxMessageSize`, the per-peer overhead rate limit, and the flag that disconnects a peer above that limit. nim-libp2p applies them inside GossipSub, and it fails node creation when a limit applies to nothing. That configuration needs the nested `Libp2pConfig.gossipsub` struct, so `flake.lock` moves the `libp2p` input to a rev that has it.

## Affected Areas

`src/topic_queues.h` and `src/topic_queues.cpp` own the backlog, its lock and its bounds. They replace three parallel maps that four translation units had to keep in lockstep.

`src/callbacks.cpp`, `src/gossipsub.cpp` and `src/plugin.cpp` push, pop, release on unsubscribe, and release on stop through that type. `src/plugin.cpp` also maps the ingress keys onto `Libp2pConfig.gossipsub`.

`src/config.h` adds the six new keys and a `parseNonNegative` helper. The helper rejects a negative value, a float, and a value above the target type. A negative queue bound read straight into `size_t` wraps into a huge positive one, and nim-libp2p refuses a negative ingress limit.

`src/metrics.cpp` exports `libp2p_module_gossipsub_queue_depth` and `libp2p_module_gossipsub_queue_dropped_total`, both labelled by topic, so a drop is visible rather than silent.

`.github/workflows/ci.yml` marks the Cachix step `continue-on-error`. The cache is an optimization, and a 403 on the push must not fail a job that built and tested everything.

Tests: a unit suite for `TopicQueues` that needs no live node, plus integration coverage of both bounds, of the disabled queue, and of the ingress limits. Docs: `README.md`, `metadata.json`, `config.example.json`, tutorials 7 and 8.

## Impact on Library Users

| Key | Default | Notes |
| --- | --- | --- |
| `gossipsubQueueMaxMessages` | `1024` | Messages held per topic. `0` disables the backlog. |
| `gossipsubQueueMaxBytes` | `4194304` | Bytes held per topic. `0` disables the backlog. |
| `gossipsubMaxMessageSize` | `0` | Largest message accepted or sent. `0` keeps the core 1 MiB limit. |
| `gossipsubOverheadRateLimitBytes` | `0` | Per-peer budget of protocol-overhead bytes per interval. `0` disables the limit. |
| `gossipsubOverheadRateLimitIntervalMs` | `0` | Refill interval of that budget. |
| `gossipsubDisconnectPeerAboveRateLimit` | `false` | Disconnect a peer that spends its budget. |

Behaviour changes only for a consumer that lets one topic exceed a bound. Those messages were retained before, and now they are dropped and counted. The byte bound holds on an empty queue too, so keep `gossipsubQueueMaxBytes` above `gossipsubMaxMessageSize`: a larger message never fits and is always dropped. A consumer that reads only the `gossipsubMessage` event should set `gossipsubQueueMaxBytes` to 0. No API signature changes, and every key is optional.

An invalid value now throws instead of silently wrapping. `load()` catches it and falls back to the defaults, as it already did for a wrong-typed field.

## Risk Assessment

Nothing in the mesh, validation or relay path changes, so network behaviour stays the same. The enqueue path gains two comparisons inside the lock it already took. A scrape samples the counters under that lock and formats its series outside it, so a node with many topics does not stall the Nim dispatch thread while Prometheus reads.

The drop counter survives an unsubscribe, because a reset Prometheus counter reads as a target restart. A topic that dropped nothing loses its map entry on release, so a node that cycles topics does not grow one entry and two series per topic forever.

`stop()` releases the backlog only when `libp2p_ctx_stop` succeeds, since a node that failed to stop still delivers. Known gap, unchanged here: nothing proves the nim-ffi listener has quiesced when `stop()` returns. A late delivery repopulates the map. The map is mutex-guarded, so the worst case is a stale message after a restart. That ordering guarantee belongs in the nim-ffi teardown path.

The `flake.lock` bump pulls in every nim-libp2p change between 91abb9c and 01068c8. CI covers that surface.

## References

The bound follows `Subscription.ch` in go-libp2p-pubsub (`topic.go`, `WithBufferSize`) and the per-peer bounds in nim-libp2p (`libp2p/protocols/pubsub/pubsubpeer.nim`).


src/topic_queues.h
richard-ramos, line 1: Is it strictly required for this header to have the code? or could it be splitted into topic_queues.h and topic_queues.cpp ?

src/topic_queues.h
```diff
@@ -0,0 +1,114 @@
...
+        std::lock_guard<std::mutex> lock(m_mutex);
+        if (m_maxMessages == 0 || m_maxBytes == 0) {
+            return false;
+        }
+        auto& t = m_topics[topic];
+        if (t.messages.size() >= m_maxMessages ||
+            (!t.messages.empty() && t.bytes + payload.size() > m_maxBytes)) {
+            ++t.dropped;
+            return false;
+        }
```
richard-ramos: the `!t.messages.empty()` would cause the byte limit to be skipped for empty queues.

```suggestion
        const bool exceedsByteLimit = payload.size() > m_maxBytes || t.bytes > m_maxBytes - payload.size();
        if (t.messages.size() >= m_maxMessages || exceedsByteLimit) {
            ++t.dropped;
            return false;
        }
```

richard-ramos (approved):

PR #103, opened 2026-08-19
feat(bridge): restore the inbound accept queue, report the peer id, add pingPeer

gmelodie:
## What

Makes the custom-protocol bridge usable by another module that runs its own protocol on this node instead of starting a second libp2p node. First consumer: logos-co/logos-delivery-module#89.

## Why the accept path did not work

PR #82 added the inbound accept queue. The nim-ffi cbind migration (#77) rewrote `onIncomingStream` to only emit an event, so the queue lost its only writer and `protocolAcceptStream` always timed out. Nothing caught it, because `tests/CMakeLists.txt` skips every integration test when `lib/libp2p.so` is absent, and that is what CI does on every run.

## Changes

- `InboundStreamQueues` (`src/stream_queues.{h,cpp}`) holds inbound streams per protocol, in the shape `TopicQueues` already uses. The inbound-stream event enqueues `{streamId, peerId}` again.
- The queue holds 1024 streams per protocol. Past that the newest stream is dropped, released without a wait (the callback runs on the Nim dispatch thread) and counted in `libp2p_module_protocol_stream_dropped_total`, next to a `libp2p_module_protocol_stream_queue_depth` gauge.
- `protocolAcceptStream` returns `{streamId, proto, peerId}`. A service protocol answers the peer that opened the stream, so it needs to know which peer that is.
- `mountProtocol` no longer needs `emitEvent`. The module boundary carries no push events, so a consumer in another module can only poll, and the queue is what keeps the stream reachable.
- `streamRelease` takes the stream out of the queue, and `stop` frees the queues while the drop counters survive.
- New `pingPeer(peerId, timeoutMs)` returns `{peerId, rttMs}` over `/ipfs/ping/1.0.0`. The bridge had no way to check that a peer is still alive.
- `queueSeries()` (`src/metric.h`) formats the depth and drop series for both poll queues, so `TopicQueues` and `InboundStreamQueues` share one copy of that loop instead of two.
- `callAsync()` (`src/plugin.h`) holds the fire-and-forget half of the promise dance that the over-cap release path needs, next to the awaiting half that owns the same rule about who deletes the promise.

## Dependency

`libp2p` is pinned to `nim-libp2p` master, which carries `peerId` on `IncomingStreamEvent` since vacp2p/nim-libp2p#2969 (merged as `c7bed9be`).

## Tests

`tests/unit_stream_queues.cpp` covers order, per-protocol separation, the cap and its drop counter, removal, and the counter surviving a release. It links without `libp2p.so`, so CI runs it.

The two accept tests now assert the peer id, and `protocol_bridge_ping_peer` and `protocol_bridge_ping_unknown_peer_fails` are new. CI does not run that layer: `nix flake check` builds the test tree with no `lib/libp2p.so` and prints `[Libp2pTests] libp2p not found in ../lib — skipping integration tests`. Run it locally with a real `libp2p.so` staged in `lib/`.

## Known limits

- An accept pump waiting in `protocolAcceptStream` when the node stops waits out its own timeout, and a restart of this module invalidates every `streamId`.
- The over-cap release path calls into `ctx` from the Nim dispatch thread, while `destroyContext()` frees `ctx` from the module thread. Nothing orders the two, so a stop that races an inbound stream can reach a freed context. Every other call in the wrapper carries the same unguarded `if (!ctx)` check, so a context lifetime lock is its own change. #107 is that change, stacked on this branch: it guards `callAsync` with the same lock and drops the bare check in `releaseStreamNoWait`, which closes this path.
- A full queue drops the stream and emits no `protocolStream` event, so an event-driven consumer that stops releasing streams loses them with only the drop counter to say so. Drive a protocol from the queue or from the event, never from both.

### DIFF
--- a/src/protocol_bridge.cpp
+++ b/src/protocol_bridge.cpp
@@ -13,9 +13,13 @@ constexpr char kPingCodec[] = "/ipfs/ping/1.0.0";
 constexpr size_t kPingSize = 32;
 
 std::string randomPingPayload() {
+    static_assert(kPingSize % sizeof(std::random_device::result_type) == 0);
     std::random_device rd;
     std::string out(kPingSize, '\0');
-    for (auto& c : out) c = static_cast<char>(rd() & 0xff);
+    for (size_t i = 0; i < kPingSize; i += sizeof(std::random_device::result_type)) {
+        const auto draw = rd();
+        std::memcpy(out.data() + i, &draw, sizeof(draw));
+    }
     return out;
 }
 
@@ -89,8 +93,7 @@ StdLogosResult Libp2pModuleImpl::protocolRequest(const std::string& argsJson) {
 
     auto d = dial(peerId, proto);
     if (!d.success) return {false, {}, "protocolRequest: dial failed: " + d.error};
-    uint64_t streamId = 0;
-    try { streamId = d.value.get<uint64_t>(); } catch (...) {}
+    const uint64_t streamId = asStreamId(d.value);
     if (streamId == 0) return {false, {}, "protocolRequest: dial returned no stream"};
 
     auto w = streamWriteLp(streamId, requestBytes);
@@ -214,8 +217,7 @@ StdLogosResult Libp2pModuleImpl::protocolAcceptStream(const std::string& argsJso
 StdLogosResult Libp2pModuleImpl::pingPeer(const std::string& peerId, int64_t timeoutMs) {
     auto d = dial(peerId, kPingCodec);
     if (!d.success) return {false, {}, "pingPeer: dial failed: " + d.error};
-    uint64_t streamId = 0;
-    try { streamId = d.value.get<uint64_t>(); } catch (...) {}
+    const uint64_t streamId = asStreamId(d.value);
     if (streamId == 0) return {false, {}, "pingPeer: dial returned no stream"};
 
     const std::string payload = randomPingPayload();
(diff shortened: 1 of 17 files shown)


src/protocol_bridge.cpp
```diff
@@ -188,24 +204,55 @@ StdLogosResult Libp2pModuleImpl::protocolAcceptStream(const std::string& argsJso
...
     }
-    auto it = m_inboundStreamQueues.find(proto);
-    if (it == m_inboundStreamQueues.end() || it->second.empty()) {
-        return {false, {}, "no inbound stream"};
+
+    StreamReadExactlyRequest readReq{};
+    readReq.streamId = streamId;
+    readReq.numBytes = static_cast<int64_t>(kPingSize);
+    std::vector<uint8_t> echo;
+    auto r = callSyncWith("pingPeer: read failed",
```
richard-ramos: so, looking at this code, i see that we do have a timeout that could possibly leave the `readExactly` running. This we cannot control from c++ since the `readExactly` future exist on nim's side.

I'll look into adding cancelation mechanics to the bindings. That sounds fun

richard-ramos (approved):
> The over-cap release path calls into ctx from the Nim dispatch thread, while destroyContext() frees ctx from the module thread. Nothing orders the two, so a stop that races an inbound stream can reach a freed context. Every other call in the wrapper carries the same unguarded if (!ctx) check, so a context lifetime lock is its own change.

Is this to be handled on a future PR?

gmelodie:
> Is this to be handled on a future PR?

yes! https://github.com/logos-co/logos-libp2p-module/pull/107

PR #107, opened 2026-08-20
fix(ctx): guard the libp2p context pointer against a concurrent destroy

gmelodie:
## Summary

`Libp2pModuleImpl::ctx` was a plain `LibP2PCtx*` that nothing synchronized. Every host-facing call tested it for null and then passed it to a binding as two separate steps, so a `destroyContext()` that landed between the check and the submit was invisible to the check: the reader handed a pointer that `libp2p_ctx_destroy` had already freed to the Nim side. The pointer itself was also a plain data race, written by one thread and read by others with no atomic and no lock.

This PR gives the pointer the same shape of guard `emitEvent` already has through `m_emitEventLock`. A new `m_ctxLock` (`std::shared_mutex`) covers `ctx`: readers hold it shared across the null check *and* the submit into libp2p, and `createContext`/`destroyContext` hold it unique. The lock never covers an await, and `libp2p_ctx_destroy` runs outside it, because the teardown waits on Nim workers that may themselves be blocked on the lock.

Two neighbours of the same bug came with it. `applyOptions` rebuilds the buffers `m_libp2pConfig`'s `NimFfiStr`/seq views borrow, and `libp2p_ctx_create` reads those views for the length of a 5 s await, so a second `createNode` could free the array under an in-flight create; the create path now runs under `m_createLock`. And both event listeners run on the Nim dispatch thread against a bare `this`, so `~Libp2pModuleImpl` was free to run while one of them was inside `m_topicQueues` or `emitEventSafe`; they now receive a heap `ListenerGate` as their `user_data` instead of `this`, hold its lock shared for their whole body, and reach the module through `gate->owner`. `destroyContext` takes that lock unique and nulls the owner before it tears the context down. The wait carries no deadline, because a deadline that expires still frees a module a listener is inside.

## Affected Areas

- `src/plugin.h`: `m_ctxLock` guards `ctx`. `callSyncWith` owns the whole submit window (check, `invoke`, release) and the promise/await tail moved into a shared `awaitReply`, so `callStaticWith` keeps its ctx-free path instead of being borrowed for the ctx path. `callAsync` stops being static and takes the same shared lock, which covers the fire-and-forget submits from the dispatch thread.
- `src/plugin.cpp`: `createContext` installs the pointer and registers the listeners under the unique lock, against a fresh `ListenerGate`. `destroyContext` detaches the listeners first, nulls the pointer under the unique lock, then tears down outside it. `ensureContext()` serializes `applyOptions` and every create under `m_createLock`. `hasCtx()` replaces the raw reads in `ok`, `status` and `createNode`.
- `src/gossipsub.cpp`, `src/custom_handlers.cpp`: the pre-checks that gate `publishEmitEvent` read through `hasCtx()`.
- `src/callbacks.cpp`: both listeners take the gate lock shared for their body and return when the owner is null.
- `src/stream.cpp`: `releaseStreamNoWait` drops its own `if (!ctx) return;`. The guarded `callAsync` now carries that check.
- `tests/sync.cpp`: `sync_reader_races_context_create` hammers `peerInfo()` from a second thread, with a handshake so the loop is provably live while `start()` publishes the pointer. `sync_stop_without_a_context_is_rejected` pins the `stop()` behaviour change.

## The hazard from #103

#103 lists it under its known limits: the over-cap release path calls `libp2p_ctx_stream_release` from `onIncomingStream` on the Nim dispatch thread, through `releaseStreamNoWait` and `callAsync`, while `destroyContext()` frees `ctx` from the module thread. This PR closes it. `callAsync` holds `m_ctxLock` shared across its own null check and its submit, `releaseStreamNoWait` no longer checks on its own, and the gate keeps the listener that calls it from running against a freed module.

## Impact on Library Users

No API change. Three behaviour changes:

- `stop()` used to read `ctx` with no check at all and pass it straight to `libp2p_ctx_stop`. It now routes through the guarded `callSync`, so calling it without a context returns `"No libp2p context"`.
- `createNode`, `start` and `newPrivateKey` serialize against each other for the length of a create (up to `kNewContextTimeoutMs`) instead of racing.
- `~Libp2pModuleImpl` blocks until a listener that is already running returns. The wait has no deadline. A teardown that reports failure leaks one `ListenerGate` (a mutex and a pointer) instead of freeing a module the Nim event thread still points at.

The cost on the hot path is one uncontended shared lock per FFI call.

## Risk Assessment

- The guard covers the submit only. An operation already in flight when the context is destroyed still depends on the Nim side failing its callbacks cleanly; this narrows the window rather than removing it, which a `shared_ptr`-owned context would.
- A listener the Nim side dispatches *after* a failed `libp2p_ctx_destroy` finds a null owner and returns. The price is one leaked gate per failed teardown. That is the case `src/plugin.cpp` already documents above `destroyContextChecked`: a non-OK teardown means the Nim workers may still be live.
- Lock order is one way (gate lock, then ctx lock; create lock, then ctx lock) and nothing holds a lock across an await, so the teardown cannot deadlock against a worker. `detachListeners()` releases the gate lock before `destroyContext` takes `m_ctxLock`, and the teardown itself runs with neither held.
- No CI job exercises any of this. `tests/CMakeLists.txt` skips the whole integration target when `lib/libp2p.so` is absent, which is what every job does, and the sanitizer jobs build the same unit-only target. The new tests need a staged `libp2p.so` to run at all, so TSAN in CI will not report the race they pin.

## References

- Issue: `ctx` has no lifetime guard, found while reviewing #103.
- Base is #103's branch, `feat/delivery-support`. Merge #103 first.

### DIFF
--- a/src/plugin.cpp
+++ b/src/plugin.cpp
@@ -211,15 +211,25 @@ StdLogosResult Libp2pModuleImpl::createContext() {
         return {false, {}, m_initError};
     }
 
+    std::unique_lock<std::shared_mutex> lock(m_ctxLock);
     ctx = r.newCtx;
 
-    // Register listeners before start so incoming protocol streams and pubsub
-    // messages are delivered once the node is running. Destroying the context
-    // frees the listener boxes.
+    // Registered before start so incoming protocol streams and pubsub messages are delivered once the node runs; destroying the context frees the listener boxes.
     libp2p_ctx_add_on_incoming_stream_listener(ctx, &Libp2pModuleImpl::onIncomingStream, this);
     libp2p_ctx_add_on_pubsub_message_listener(ctx, &Libp2pModuleImpl::onPubsubMessage, this);
 
     return {true, {}, ""};
+}
+
+bool Libp2pModuleImpl::hasCtx() const {
+    std::shared_lock<std::shared_mutex> lock(m_ctxLock);
+    return ctx != nullptr;
+}
+
+StdLogosResult Libp2pModuleImpl::ensureContext() {
+    std::lock_guard<std::mutex> lock(m_createLock);
+    if (hasCtx()) return {true, {}, ""};
+    return createContext();
 }
 
 StdLogosResult Libp2pModuleImpl::createNode(const std::string& config) {
@@ -231,7 +241,8 @@ StdLogosResult Libp2pModuleImpl::createNode(const std::string& config) {
         fprintf(stderr, "libp2p_module: %s\n", msg.c_str());
         return {false, {}, msg};
     }
-    if (ctx) {
+    std::lock_guard<std::mutex> lock(m_createLock);
+    if (hasCtx()) {
         std::string msg = "createNode: node already created";
         fprintf(stderr, "libp2p_module: %s\n", msg.c_str());
         return {false, {}, msg};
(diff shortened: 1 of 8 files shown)


src/plugin.cpp
```diff
@@ -250,14 +261,31 @@ StdLogosResult Libp2pModuleImpl::setLogLevel(const std::string& level) {
     return {true, {}, ""};
 }
 
+void Libp2pModuleImpl::drainListeners() {
+    if (!m_listenerLock.try_lock_for(std::chrono::milliseconds(kDefaultOpTimeoutMs))) {
```
richard-ramos: Hopefully during those 10s, all callbacks that are running that might be using the userdata (`this`), will be finished. It's probably an edge case tho

richard-ramos (approved):

